PR #7, opened 2022-05-21
Debug and other enhancements

SensorsIot:
Hi Timm,
After these changes, I have our scenario working. We can Skype if you want.


timmbogner (approved):
I will tweak these a little and apply the changes to the two official sketches, then I'll come back and standardize these with the official ones.

PR #10, opened 2022-05-23
Moved band and SF to the default file

SensorsIot:
After all the changes it is no more necessary to have them in credentials


SensorsIot:
Not a good idea

SensorsIot:
Still not a good idea

PR #23, opened 2022-06-25
Add MQTT port selection and MQTT user/password auth support

thefeiter:
First of all: Great Project! Got it up and running with spare parts laying around in under an hour.

In my setup, I want to use an MQTT broker on my server which requires authentication, so I added a few lines to be able to set MQTT user credentials.
Also I added an option to select the port of the broker.

It works in my test setup.

Maybe it can be simplified.


timmbogner:
Yo, thanks! I appreciate the help! The dev branch has some drastic changes happening, so I think that explains the conflicts? I will have a look on Monday and possibly just pull this to the main branch while the new version is under testing/construction.

Devilbinder:
I unfortunately can't see what is causing the conflict. This code does not seem to affect the changes that I have made. Specifically left the current gateway examples alone until I have conformation that the Universal Gateway example is working.

Devilbinder:
Resolved the conflict on https://github.com/Devilbinder/Farm-Data-Relay-System/tree/dev2
@thefeiter you should just need to merge my branch into yours to resolve the conflict.

thefeiter:
Done, thanks!

PR #26, opened 2022-06-28
Added GeneralGPS sensor

Gulpman:
I added a GPS sensor, which will transmit longitude, latitude and altitude as data values.

I also added LATITUDE_T, LONGITUDE_T and ALTITUDE_T to the list of sensor values.

I tried to be as platform-agnostic as possible. So this should (hopefully) run on arbitrary platforms. 

As I'm not a programming guru, I'm more than open for corrections and advices. :)


timmbogner:
Haha whoops i thought i'd merged that, but on second thought i'm going to do some housekeeping first.

hjboven:
If you would forward the HDOP value as well it would make it near perfect. The HDOP Value is needed to access the quality/accuracy of the provideded coordinates.

PR #29, opened 2022-06-30
Data Pipeline

Devilbinder:
Data is no longer aggregated into a single buffer. Data is aggregated on a per peer bases.

Fix Universal_Gateway not compiling after house keeping


Devilbinder:
Please let me know if this is working. I'd like to add formal [doxygen](https://doxygen.nl/) documentation if it does.

Devilbinder:
Oh I did not see your comment on #39

timmbogner:
I'm still struggling to figure the configuration out, but I'm hoping maybe we can step back and talk about what the final configuration should look like. Last night I had a brainstorm and jotted down some ideas that are new for me, but I expect similar to what you're already doing. The main thing i've tried to work out is the configuration. What could we do to be able to define a function within the setup() function? For example an .ino would look like:
```
#include stuff.h
FDRSGateway FDRS(yabba, dabba);
ESPNOWevent ESPNOWunreg(doo);
UARTevent UART();
setup(){
ESPNOWunreg.action() = {UART.send(); DBG("espnow unregistered");}
loop(){
FDRS.loop();
}
```
I'm on mobile, so that's just a quick look at my concept. I can elaborate further, but I think referring to each object as an 'event' will make it expandable in the way I envision. How and where to define the action functions is another question, but I think it will help us make a user configuration setup that is intuitive enough for beginners.
I'm also debating using "unregistered" instead of "unknown".

Gulpman:
> Here is the most simple example I can think of to pipe data from A -> B. This pipes all data from ESPNOW to MQTT.
> 
> My Idea is to create a examples for each permutation. LoRa to serial, ESPNOW to serial, LoRa to ESPNOW etc. Only shoved everything into one example for testing purposes.

Thank you very much for the example. But I'm struggling with a LoRa-> Serial example... 
@Devilbinder: Can you give me an example which is receiving LoRa Data and piping it to Serial?

And as I tried to find out how to do that: Is there a possibility to pipe LoRa -> Serial & LoRa (so I have the debug output and can forward?)

Sorry for not being a pro, but a working piece of code would help me in testing further.

Issue #35, opened 2022-07-01
Proposal: Move data types to a separate file within FDRS_install


Gulpman:
I was wondering if it wouldn't be easier for maintainance and cleaner code if the data types for sensors would go to a separate file within the FDRS_Install folder, e.g.: 

FDRS_Install\FDRS_datatypes.h
```
#define STATUS_T    0  // Status 
#define TEMP_T      1  // Temperature 
#define TEMP2_T     2  // Temperature #2
#define HUMIDITY_T  3  // Relative Humidity 
#define PRESSURE_T  4  // Atmospheric Pressure 
#define LIGHT_T     5  // Light (lux) 
[...]

```
This way it could be easily included within the sensor sketches by using 
`#include <FDRS_datatypes.h>`

This would shorten the fdrs_sensor.h's for each sensor a little bit. But the great advantage would be that if the data types are extended it does not have to be edited in every single sensor sketch to match the newest version. Only updating the <FDRS_datatypes.h> file would be sufficient.

Or am I overseeing something?

thefeiter:
I guess this would also be a good Idea for the fdrs_functions and fdrs_sensor files.
Can one make a library out of this repo which one could install in the libraries folder?

thefeiter:
I would suggest moving all the files containing the same code to a central directory which is to be placed in the libraries folder.

timmbogner:
The types as globals is also something that Andreas pushed for as well. My reluctance was because I thought those should only be changed project-wide. That is, I wanted people to contact me to add new types, that way we didn't end up with a bunch of conflicting new ones.  The other issue that I always stuck to was that I wanted the system to stand alone. Like, you can run it out of its sketch folder with no installation. I was happy to have someone see the repo, let alone move files around just to try it out.

Soon, the library approach will make this a non-issue, as things will be installed as the norm. Unless anyone objects. For now, I think I could have it say:

```
   #ifdef USE_GLOBALS
   #include <library.h>
   #else
   #include "library.h" 
   #endif
```
This would choose the header in the project folder if globals is not available.  Seem good?

timmbogner:
I forgot to wrap up my initial thought, which was that the "FDRS_Install\FDRS_datatypes.h" idea seems like a good one.

Devilbinder:
In favor of moving any duplicate code to the root of the project. Was my original idea with #17 and #20.
It will make code maintenance much easier.

[...]

PR #37, opened 2022-07-02
Unify common files via script

thefeiter:
I wrote a small script which copies the `fdrs_functions.h` and `fdrs_sensors.h` from the FDRS_... directories into the known directoreis where these files are needed.

With this one has to only update/change the files located in `FDRS_Sensor` and `FDRS_Gateway` and can then automatically overwite all the other files.

This is to be used as a developer tool but looks like a good step inbetween copying everything by hand and having only one file in the libraries directory like mentioned in #35 

(I have no idea, how some of the files look like completely changed to git...)


timmbogner:
@thefeiter btw I really appreciate this, I just left it out of the latest revision because it's no longer an issue.

thefeiter:
No problem! Was just a workaround :+1:

PR #42, opened 2022-07-03
Add a logging function to write incoming Packets to SD card

thefeiter:
This came to my mind when my internet connection was interrupted and data was not sent to my mqtt broker.

SD cards work on SPI, so there has to be another Chip/slave select pin, but it is working good, since the Master does only talk to one slave at once (it is either using LoRa or the SD card)

For now I am just appending the JSON data to a hardcoded file whenever it is being sent over Serial.

This is meant to be able to retrieve sensor values when there is an Error with your internet connection or WiFi

TODO:

- [x] Choose Filename via config
- [x] Better datastructure instead of JSON -> CSV
- [x] choose filename on function call
- [x] Some Documentation
- [x] Somehow implement a better timestamp method

After Merge:

- [x] start working on internal flash option
- [ ] sending the missed readings with timestamps


thefeiter:
Any Ideas on getting some kind of timestamp without WiFi/NTP?

thefeiter:
Whit this timestamp method the 1/10 seconds since reset are counted (based on millis()). This shold roll over after 4971 hours, this should be sufficient to retrieve lost sensor values.

Logs look like this atm:

```
0.20 : [{"id":11,"type":16,"data":3.265350819}]
4.70 : [{"id":2,"type":1,"data":26.25}]
8.10 : [{"id":4,"type":1,"data":28.5625}]
10.20 : [{"id":11,"type":16,"data":3.264254332}]
15.20 : [{"id":2,"type":1,"data":26.25}]
18.80 : [{"id":4,"type":1,"data":28.625}]
20.30 : [{"id":11,"type":16,"data":3.265350819}]
30.30 : [{"id":11,"type":16,"data":3.266447306}]
36.10 : [{"id":2,"type":1,"data":26.25}]
40.30 : [{"id":4100,"type":144,"data":28.625}]
40.30 : [{"id":11,"type":16,"data":3.266447306}]
46.60 : [{"id":2,"type":1,"data":26.25}]
50.40 : [{"id":11,"type":16,"data":3.264254332}]
60.40 : [{"id":11,"type":16,"data":3.259868383}]
67.60 : [{"id":2,"type":1,"data":26.25}]
70.40 : [{"id":11,"type":16,"data":3.266447306}]
```

thefeiter:
> If we add this, I think it would be good to have it only run if MQTT is down.

Yep, so this would only be used in an MQTT gateway.



> Also, can you establish the filename when calling the function?

Good idea!



> It may also be a good idea to implement writing to other filesystems as well like internal flash

For internal Flash I would have do do some reading, this is new to me. But it would be much easier since you don't need another SD card breakout board.





> CSV might be a more portable format for the data

Good, it is at least easier to use than json.



> Maybe somewhere farther down the line we can add a method of sending the missed readings with timestamps over MQTT when a connection is reestablished, but there are several other features that I want to implement first.

Maybe syncing an internal clock when internet is available and using this as a timestamp for when internet is down.

thefeiter:
I changed a lot in this commit
- The reconnect function now takes 2 arguments: attempts to reconnect and if debugging output shall be silent
- In the loop the MQTT client tries to reconnect if it is not connected and loops afterwards (even without connection)
- Reconnecting MQTT does not block the thread forever
- I added a `mqtt_publish() `to remove redundand t code in `sendMQTT` and `releaseMQTT`, here if the MQTT publish fails, it logs to SD

thefeiter:
Improvement ideas appreciated.

thefeiter:
This seems to be the best option:
If wifi is activated, we keep time by using an NTPClient which also keeps time when the connection is interrupted via millis and is able to return a Unix epoch timestamp.
If wifi is off, it falls back to my original millis timekeeping.

I reduced the resolution to seconds to produce similar timestamps like the NTPClient.

thefeiter:
From my side, I am ready for merge.

FDRS_Gateway/FDRS_Gateway.ino
```diff
@@ -132,12 +130,20 @@ void loop() {
...
+  #ifdef USE_WIFI
   if (!client.connected()) {
-    DBG("Connecting MQTT...");
-    reconnect();
+    reconnect(1, true);
   }
-  client.loop();
+  client.loop(); // for recieving incoming messages and maintaining connection
+
+  timeClient.update();  //update internal clock if possible
```
timmbogner, line 139: Is this updating every loop? Should we maybe just update the time when it connects?

thefeiter:
The time client tries to update every 60 seconds only. (internally) If needed one can adjust this in the setup.
Updating once might be too inaccurate because of the possible drift of millis().

08.07.2022 04:26:19 Timm Bogner ***@***.***>:

PR #44, opened 2022-07-05
Dev - cleaned up library structure and got rid of redundant files

Gulpman:
I altered the library structure, so examples are directly shown when you open the example menu in Arduino IDE.

- addes keywords.txt (not complete) and library.properties with project information (please complement desciption)
- renamed Examples to examples to be displayed correctly within the Arduino IDE
- removed local occurances of fdrs_sensor.h and fdrs_functions.h
- moved obsolete FDRS_Sensor and FDRS_Gateway to resources folder. There is still documentation in them which should not get lost. 
- moved Sensors folder to resources folder - sensor sketches need to be refactored and should be moved to a subfolder of examples as sensor examples in a later step.
- For easier debugging I addes some random values to 1_Lora_Sensor and 2_ESPNOW_Sensor. To better distinguish them while debugging i referenced the first digit of the tempterature value to the READING_ID.
- made global settings the default and corrected wrong usage of BAND and SF instead of FDRS_BAND and FDRS_SF.


timmbogner:
I took it to where I would prefer it.  I sort of liked the way that the examples menu showed up before, but I see how taking it "full library" is a good idea. The idea of leaving the two sketches in the main folder is to have solidly defined "official" sketches. It perhaps doesn't make a lot of sense, but my decision is also made because I like the current structure of the repo/documentation. Plus it's been like that forever and I think it makes the repo more intuitive to new users.

timmbogner:
I just noticed that the FDRS_Gateway and Sensor sketches still show up in the examples, which I like.
I'm also excited about the syntax highlighting, although it isn't showing up in the Arduino IDE. Did I miss something?

Gulpman:
> I'm also excited about the keyword highlighting, although it isn't showing up in the Arduino IDE. Did I miss something?

No you didn't. It should highlight code utilizing the library, from what I read. I have to check why it is not highlighted. I stumbled upon it while reading about how a library should be structured and found it useful. 
Will check.

timmbogner:
Excellent! I am now relieved and less confused. I see the confusion and will try to cite commits and PRs like I see other people doing.
So just to reiterate: I kind of like the fact that the FDRS_Gateway/Sensor sketches still show up in the examples menu, so no need to fuss about that.
@Devilbinder Do you have an opinion on putting the common files in a "src" folder vs the root? I can see value in both options.

PR #45, opened 2022-07-06
Create internal flash (LittleFS) logging functionality

thefeiter:
As discussed in #42 I implemented a possibility to log data to the internal SPIFFS Filesystem.


thefeiter:
I have checked the date in the filesystem via a sketch. Seems to be working. But there is no convenience method to load the data onto your computer without uploading a different sketch.

A Http Fileserver maybe...

thefeiter:
Works as expected.
Retrieving the files shall be done sometime later.

timmbogner:
Okedoke. A few thoughts as I contemplate this:
- A major usefulness of this will be to automatically update the server with missed readings on reconnect.  Especially since the filesystem isn't super easy to get files from, so let's leave it here for now. In the future there might be more we can do, like triggering the gateway to send the entire file.
- I think the seconds as a back up isn't necessary. If it can't give us real timestamps then it shouldn't bother.
- If I remember right, it's best to keep flash writes to a minimum, as rapid writes can corrupt the card.  It might be better to buffer writes and make several at once. I'm not sure about this; its something to research. Will an SD card or flash hold up to a massive influx of individual readings? Have you tried running it against the Stress Test? I'm still looking for my ESP-CAM to test with.

thefeiter:
> * A major usefulness of this will be to automatically update the server with missed readings on reconnect.

True



> * I think the seconds as a back up isn't necessary.

For the internal Flash, yes. Since one cannot easily read the files. But for the SD log it is still useful to be able to see not only in which order values came in but also how far apart time wise. And since I would keep the seconds since reset for the SD log I would also levae them in the Flash log.



> * its something to research

True

thefeiter:
When logging to SD+FS with the stresstest only every 7th reading makes it to the sd card

When logging only to SD only every 2nd reading comes through

When logging only to FS every 8th reading makes it to the internal flash

Sooo, uhm yeah. Maybe buffering is a good idea. But I think the stresstest is going to extremes anyways, so for normal usage this should be ok for now.

timmbogner:
> But I think the stresstest is going to extremes anyways, so for normal usage this should be ok for now. 

True.

thefeiter:
Now it is at least buffering the whole line to write. 
~97% make it to the sd card with the stresstest.
Writing to flash is now at 20%, this seems to be a weird bottleneck.
I thought the internal FS would be the fastest...

thefeiter:
https://github.com/esp8266/Arduino/issues/7271

Devilbinder:
Thanks for switching this to LittleFS. SPIFFS is a fragile headache of a file system.

PR #53, opened 2022-07-08
DEV: Refactored DEBUG and DBG to FDRS_DEBUG and FDRS_DBG

Gulpman:
As mentioned in #47 the names DEBUG and DBG prevent the project to compile on PlatformIO.
It also will prevent issues when other libraries will be used in the project as it is likely that DEBUG and DBG may be used there as well.

In addition I removed a duplicate entry from keywords.txt

### DIFF
--- a/FDRS_Gateway/README.md
+++ b/FDRS_Gateway/README.md
@@ -16,7 +16,7 @@ In this example, the gateway is set to take any ESP-NOW packet it receives and s
 ## Options
 ### ```#define UNIT_MAC (0xNN)```
 The UNIT_MAC is the ESP-NOW and LoRa address of the gateway. This is the address that nodes or other gateways will use to pass data to this device.
-### ```#define DEBUG```
+### ```#define FDRS_DEBUG```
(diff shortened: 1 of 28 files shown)


timmbogner:
Is DBG() definitely a no-go? I really like the brevity of it and would like to keep it short.

Gulpman:
I think it is not a no-go. Im my opinion it is more about consistency. Therefore I would prefer switching to FDRS_DBG.

Devilbinder:
Common practice to use a project prefix with debug macros. The issue is that everyone likes using `DEBUG` so no-one can use it because it causes conflicts like we see now with #47. Also `DEBUG` is a build reserved macro in many cases.

Example of a widely used repos config system.
https://github.com/Mbed-TLS/mbedtls/blob/a343d34f0d6e11576863bcd626ac966b42f2deb7/include/mbedtls/mbedtls_config.h#L1397

timmbogner:
I support that with the full DEBUG term. It makes a lot of sense and is no issue.

What I'm dragging my feet on is the DBG() function. If DBG is commonly used, can we find a three letter term that isn't? I like it short, sorry if this is sort of petty. I have just always liked how easy it makes it to add verbosity.

I won't die on this hill, but I did want to make a stand.

PR #56, opened 2022-07-10
Buffer log writes before saving them to the file.

thefeiter:
As discussed in #45 I messed around with buffering the log writes to SD and Flash. 

This is to minimize package loss on file writes. In "normal" use it should workout fine, but when running the stresstest one can find some bottlenecks.


thefeiter:
Atm I use the same buffer for SD and flash.

Stress test is set to delay 30 ms btw.

For USE_SD_LOG only package loss is at 5% (worse than without buffering, was 3%)
For USE_FS_LOG only package loss is down to 17% (w/o buffer it was 80%) 

Now I am thinking of using the buffer only for the LittleFS logging, since the SD logging package loss got worse.
But this would add more complexity and redundancy to the code.
What do you think @timmbogner ?

timmbogner:
Sorry, I think I over-emphasized the importance of the stress test for your functionality. I have to look a little more when I'm at the computer, but it looks great! My biggest concern is that rapidly opening/closing a file on an SD could corrupt the filesystem of the card. That was my main thought behind buffering writes.

thefeiter:
Awesome, then the buffering would be perfect to reduce rapid writes!

10.07.2022 20:18:32 Timm Bogner ***@***.***>:

FDRS_Gateway/FDRS_Gateway.ino
```diff
@@ -145,13 +145,13 @@ void loop() {
...
   timeClient.update();  //update internal clock if possible
   #endif
-#ifdef USE_SD_LOG
-unsigned long current_millis = millis();
-if(current_millis-last_millis >= 1000){
-  seconds_since_reset+=(current_millis-last_millis)/1000;
-  last_millis=current_millis;
-}
-#endif
+  #ifndef USE_WIFI
```
timmbogner, line 148: Can you make this also only happen when logging is enabled? Pardon if i'm misreading.

timmbogner:
This all looks really good, and I think it may work a general-use function in the future. One thing that I realized: since unix timestamps are just the number seconds since an arbitrary time/date, 'seconds_since_reset' is actually in the same format as the unix timestamp, yes? 🤯 Nice.

FDRS_Gateway/fdrs_functions.h
```diff
@@ -191,37 +196,46 @@ void getSerial() {
...
+    #ifdef USE_WIFI
+    sprintf(linebuf, "%ld,%d,%d,%g\r\n", timeClient.getEpochTime(), theData[i].id, theData[i].t, theData[i].d);
+    #else
+    sprintf(linebuf, "%ld,%d,%d,%g\r\n", seconds_since_reset, theData[i].id, theData[i].t, theData[i].d);
+    #endif
+
+    if (logBufferPos+strlen(linebuf) >= (sizeof(logBuffer)/sizeof(char))) // if buffer would overflow, release first
+    {
+      releaseLogBuffer();
+    }
```
timmbogner, line 234: Is it only released when the buffer is full? Could there be unreleased data in the buffer when MQTT reconnects that never gets out?

FDRS_Gateway/FDRS_Gateway.ino
thefeiter: Yep, that makes sense!

FDRS_Gateway/fdrs_functions.h
thefeiter: Atm, this is the case. I was thinking of also releasing the buffer regularly in the loop when it has something in it.

thefeiter:
To prevent issues with the millis() rollover I suggest changing the buffer delay functionality to compare durations instead of timestamps (https://arduino.stackexchange.com/a/12588)

thefeiter:
I set the buffer release timer to 10 seconds. I think it could be longer.

timmbogner:
I'm a dum-dum and pulled the requests out of order again, but I think I fixed it!

PR #63, opened 2022-07-13
Add ability for debug output to not have a newline or debug with newline

aviateur17:
Changed DBG -> DBGLN and added new #define DBG that does not generate a newline.   Gives developers ability to output different types of data on the same line instead of generating a new line each time.

### DIFF
--- a/FDRS_Gateway/FDRS_Gateway.ino
+++ b/FDRS_Gateway/FDRS_Gateway.ino
@@ -45,7 +45,7 @@ void setup() {
   Serial.begin(115200);
   UART_IF.begin(115200, SERIAL_8N1, RXD2, TXD2);
 #endif
-  DBG("Address:" + String (UNIT_MAC, HEX));
+  DGBLN("Address:" + String (UNIT_MAC, HEX));
 #ifdef USE_LED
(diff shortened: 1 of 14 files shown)


aviateur17:
I guess this really isn't needed.  Can use existing DBG with concatenation and String().

Example:
DBG("Incoming LoRa. RSSI: " + String(LoRa.packetRssi()) + " SNR: " + String(LoRa.packetSnr()));

I hadn't reviewed everything in the project before making this change.   I actually prefer to use printf() or ESP LOGx() functions but that would be a big change so I don't suggest that at this point.  We can probably reject this PR.

timmbogner:
I agree that it's not needed currently. The spirit of DBG() is to not obscure the code very much, but this change would encourage more lines of debug code to be added. 
It may be worth looking into whether the String() concatenation method is the best one. I started using it a while ago, but don't often see it used elsewhere. I think I saw someone using sprintf(), which is maybe more dev-approved.

aviateur17:
Probably not high priority.  When I write code for microprocessors I always have memory conservation in mind.  String class seems to use up a lot of memory (probably not a concern with ESP32) and has had reports of memory leaks.   Therefore I try to use what Espressif has provided LOGx() or C printf().   String() does make manipulation easy though.

Issue #54, opened 2022-07-09
[NEW FEATURE] Packet Confirmation and System Packets


timmbogner:
**First off, I want to extend my deep appreciation to everyone who has helped me with this project.** I'm so glad that everyone is having as much fun with FDRS as I am!
### The first of the goals that Andreas mentioned for FDRS is _**packet confirmation**_.

Packet confirmation is already done within the ESP-NOW protocol to some degree, and you can use the send callback to confirm whether or not it was delivered. From what I read, the protocol actually does some quick retries if no confirmation is received. However, because we will want to keep our methods uniform with LoRa, which has no confirmation, I propose we create a new protocol for confirming packets on both LoRa and ESP-NOW.

Something I hold sacred in FDRS is that the messages are sent as either a single object or an array of objects (DataReadings). Because this is the only expected packet format, adding another packet type brings on the question: how does the gateway know which is which? My favorite solution: the system packets are smaller, and thus if LoRa or ESP-NOW receive a packet smaller than a DataReading, it knows that it is dealing with a system packet. This of course limits you to single system packets at a time. A DataReading is 7 bytes, thus I propose the new packet be **either 6 or 4**. The packet will already know its size and sender, so it doesn't need to convey either of those.

The system packet will be sent after a message is received by a device. It will send some kind of token back to the sender that proves it received the message correctly. The sending device will be waiting (non-blockingly) to receive that proof, and retrying at intervals if not.

So the main question: **what is the proof?** I'm hoping someone can just tell me whether to use **CRC or checksum** for the token.

This may also be a good time to have a discussion about encryption/security, which will be another use of the new packet. This will not be pertinent until two-way communication is ironed out, but it's good to keep in mind. The system packet will also be used by sensors for pinging and pairing with gateways, as well as perhaps get/setting configuration parameters.

**_All thoughts, ideas, and proposals are welcome and appreciated!_**

Devilbinder:
For data acknowledgement.
A message type data field will be required. To tell the device how interpret the data. A message ID will also be required. Some simple counter will do. You don't need "proof". All your device need to know is: I sent a message with ID x and have I gotten a response about that ID.
Another approach is to have the final endpoint query the sensors but this does not scale to large networks.

CRC and checksum are used for data integrity checks not for data acknowledgement. They get appended at the end of a data frame,

CRC are superiors with the tradeoff being that is more computationally expensive unless you are using hardware accelerators. 

Here is how to do a CRC.
[...]

PR #67, opened 2022-07-16
Initial work on LoRa ACKs

aviateur17:
These are inital changes to the dev branch in order for LoRa to support ACK packets.   It is not yet totally working.  Sender and Receiver CRCs do not match and there is some work to be done to test bad CRC and adding ability to disable ACK replies on sensor side.

I've tested my code in Arduino 1.8.19 both the FDRS_Sensor and FDRS_Gateway and it should compile in both.

Todo:
Figure out why Sender and receiver CRC does not match
Confirm sensor receives ACK from gateway
Test scenario where CRC does not match
Add sensor functionality to disable ACKs from gateway

### DIFF
--- a/fdrs_functions.h
+++ b/fdrs_functions.h
@@ -284,7 +284,7 @@ void getLoRa() {
     packetCRC = ((packet[packetSize - 2] << 8) | packet[packetSize - 1]);
     memcpy(&destMAC, &packet[0], 2);              // Copy destination address to variable
     memcpy(&sourceMAC, &packet[2], 2);             //Split off source address portion of packet (2 bytes, bytes 3 and 4)
-    DBGLN("Incoming LoRa. Size: " + String(packetSize) + " Bytes, RSSI: " + String(LoRa.packetRssi()) + "dBi, SNR: " + String(LoRa.packetSnr()) + "dB, FreqError: " + String(LoRa.packetFrequencyError()) + "Hz, PacketCRC: 0x" + String(packetCRC,16));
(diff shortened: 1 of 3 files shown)


aviateur17:
Not sure how to associate this with #54.  I see DBGLN so I'll have to create another commit to change those back to DBG.  Ugh.

fdrs_functions.h
```diff
@@ -494,12 +553,12 @@ void releaseLoRa(uint8_t interface) {
...
           if ( j > lora_size) {
             j = 0;
-            transmitLoRa(broadcast_mac, thePacket, j);
+            transmitLoRa(&loraBroadcast, thePacket, j);
           }
           thePacket[j] = LORAGbuffer[i];
           j++;
         }
-        transmitLoRa(broadcast_mac, thePacket, j);
+        transmitLoRa(&loraBroadcast, thePacket, j);
```
timmbogner, line 561: This is a quick shot in the dark, but here you changed it to pass an address instead of the variable. That's not your issue is it? I'm still shaky with *s and &s to be honest.

aviateur17: When I changed the MAC addresses from 3 bytes to 2 bytes I changed some of the types from an array of uint8_t to a single uint16_t.   Because of that change I had to change the way the broadcast address is passed into that transmitLoRa function.  Since the address is no longer an array I had to add the Address of (&) in front of the variable in order to pass in the pointer.  I don't think it has any effect on why the data isn't matching.   I'd have to spend more time on it but there is something going on as far as the byte order - big endian vs little endian.  Looking at the output in the other thread the byte order is reversed.   But there is also something else going on.  Not sure but I'll have to look at it some other time.

timmbogner: I figured there was a good reason, I just wanted to check. Hey have a nice vacation!

PR #74, opened 2022-07-22
Clean up of #72

Gulpman:
- moved configuration checkup to external file to make sensor and gateway files more tidy (and smaller).
- for sensors the config now only is shown if the sensor does NOT return from Deep Sleep. For example when pressing the reset button or after a crash.
- for gateways the configuration is shown once and also can be forced to be shown with a reset. Code should work for ESP32 and ESP8266 - have only tested it with ESP32.
- fixed issue in fdrs_functions.h - if MQTT_AUTH was enabled but WIFI disabled, the compile crashed.

I needed to define a USE_ESPNOW in fdrs_gateway_config.h to make things easier. Actually this is needed anyway as it is not obligatory for a gateway to offer ESP-Now support. 
As the gateway code does not offer a central starting point like beginFDRS() as the sensor does, I had to implement the config check in the .ino file. This has been done only in 4_UART_Gateway so far.

### DIFF
--- a/FDRS_Gateway/FDRS_Gateway.ino
+++ b/FDRS_Gateway/FDRS_Gateway.ino
@@ -45,11 +45,25 @@ void setup() {
   Serial.begin(115200);
   UART_IF.begin(115200, SERIAL_8N1, RXD2, TXD2);
 #endif
+
   DBG("Address:" + String (UNIT_MAC, HEX));
+
+#ifdef DEBUG_NODE_CONFIG
+  // find out the reset reason
+  esp_reset_reason_t resetReason;
+  resetReason = esp_reset_reason();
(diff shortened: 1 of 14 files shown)


timmbogner:
beginFDRS() for the gateway would be a fine addition. Making ESP-NOW optional is a good idea in theory, but would be a big undertaking for not much payoff. That's the main reason I haven't done it. You're welcome to change it if it makes your life easier 😉

Gulpman:
> Making ESP-NOW optional is a good idea in theory, but would be a big undertaking for not much payoff. That's the main reason I haven't done it. You're welcome to change it if it makes your life easier 😉

Done 😊 See https://github.com/timmbogner/Farm-Data-Relay-System/pull/74/commits/ae85886e65ccaf403ded30b7c670b9a9a979ef80

PR #76, opened 2022-07-25
#54 Packet Confirmation - LoRa ACK and CRC

aviateur17:
Code submission per discussion #54

### DIFF
--- a/FDRS_Sensor/fdrs_sensor.h
+++ b/FDRS_Sensor/fdrs_sensor.h
@@ -293,13 +293,12 @@ void transmitLoRa(uint16_t* destMAC, DataReading * packet, uint8_t len) {
   pkt[2] = (LoRaAddress >> 8);
   pkt[3] = (LoRaAddress & 0x00FF);
   memcpy(&pkt[4], packet, len * sizeof(DataReading));
-#ifdef LORA_ACK
   for(int i = 0; i < (sizeof(pkt) - 2); i++) {  // Last 2 bytes are CRC so do not include them in the calculation itself
     //printf("CRC: %02X : %d\n",calcCRC, i);
     calcCRC = crc16_update(calcCRC, pkt[i]);
   }
-#else
-  calcCRC = LORA_NOACK_CRC;
+#ifndef LORA_ACK
+  calcCRC = crc16_update(calcCRC, 0xA1); // Recalculate CRC for No ACK
 #endif    // LORA_ACK
   pkt[len * sizeof(DataReading) + 4] = (calcCRC >> 8);
   pkt[len * sizeof(DataReading) + 5] = (calcCRC & 0x00FF);
--- a/FDRS_Sensor/fdrs_sensor.h
+++ b/FDRS_Sensor/fdrs_sensor.h
@@ -57,10 +57,18 @@ typedef struct __attribute__((packed)) DataReading {
 
 } DataReading;
 
+
+enum crcResult{
+  CRC_NULL,
+  CRC_OK,
+  CRC_BAD,
+} returnCRC;
+
 enum {
   cmd_clear,
   cmd_ping,
   cmd_add,
+  cmd_ack,
 };
 
 typedef struct __attribute__((packed)) SystemPacket {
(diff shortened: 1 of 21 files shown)


timmbogner:
I think I resolved the conflicts correctly. That's probably my mistake for merging another PR a bit ago.

aviateur17:
Better than me.  I'm just getting a hang of all this git stuff.   Should have another improvement on CRC soon.

FDRS_Sensor/fdrs_sensor.h
```diff
@@ -176,16 +187,143 @@ void beginFDRS() {
 
 }
 
-void transmitLoRa(uint8_t* mac, DataReading * packet, uint8_t len) {
+//  USED to get ACKs from LoRa gateway at this point.  May be used in the future to get other data
+// Return type is crcResult struct - CRC_OK, CRC_BAD, CRC_NULL.  CRC_NULL used for non-ack data
+crcResult getLoRa() {
+#ifdef USE_LORA
+  int packetSize = LoRa.parsePacket();
+  if ((packetSize - 6) % sizeof(SystemPacket) == 0 && packetSize > 0) {  // packet size should be 6 bytes plus multiple of size of SystemPacket
```
timmbogner, line 195: i hadn't thought of that! This means we can send multiple SystemPackets at once, right?

aviateur17: Yep!  Just like DataReading.

aviateur17:
Just added commit to enhance LoRa packet validation.  Instead of sending fixed CRC when sensor does not want an ACK the sensor will add a known byte to the CRC calculation in the case that it does not want the ACK and therefore including the CRC with the data.  The gateway will check the CRC without that known byte and if there is mismatch then add that known byte to the CRC calculation and if that matches will assume that the sensor does not want an ACK.   If that still does not match then the packet is not valid and will be discarded by the gateway.  That way the CRC is still calculated for packets where the sensor does not want a return ACK reply.

timmbogner:
> Just added commit to enhance LoRa packet validation. Instead of sending fixed CRC when sensor does not want an ACK the sensor will add a known byte to the CRC calculation in the case that it does not want the ACK and therefore including the CRC with the data. The gateway will check the CRC without that known byte and if there is mismatch then add that known byte to the CRC calculation and if that matches will assume that the sensor does not want an ACK. If that still does not match then the packet is not valid and will be discarded by the gateway. That way the CRC is still calculated for packets where the sensor does not want a return ACK reply.

I see the conundrum here and that's a clever way to fix it.

aviateur17:
There might be a code issue with #endif between the two fdrs_functions.h files, one in the root and one in the FDRS_Gateway folder.  I'm gonna fix that today if there is an issue.

FDRS_Sensor/fdrs_sensor.h
timmbogner: Out of the blue, I realized that I think misunderstood something here.

If we are receiving data and it could be either a DataReading or a SystemPacket, we can't rely on dividing by the size of the structure to distinguish between the two, since there could be common divisors. DataReading being 7 bytes makes that a bit less likely, but it could also change. If I'm not confused (which I could be) we will still have to send SystemPackets one at a time and check for them by seeing if they're smaller than DataReadings.

This code is fine, though.

aviateur17: Yeah, that's right, there are two types of data being sent between devices - DataReadings and SystemPackets.  At this point I believe the only way to distinguish the two different types is by the size of the packet.   If the structure of one or both of those types changes we'll have to change the code like you mentioned.   I'm not sure I know of another way to distinguish between the two other than size at this point.  An alternative would be to add another byte to each packet to distinguish between the two types.   Possibly do something with the CRC calculation but that would be starting to get confusing.   I'm open to ideas if you think of alternatives.

timmbogner: Distinguishing by size is okay, That's how I'd planned it. The question about images this morning got me thinking about a third type that could be a full 250 bytes. Now, this could currently be a distinguishing feature since the most DataReadings could take up is 245 bytes. But if DataReading was 10 bytes, then we could have an issue when there are 25 items.

So... Maybe we lower our max packet size for DataReadings to 249 and introduce a new BulkPacket that uses exactly 250. 
The bulk packet will have a reading id (or something),  and individual frame numbers for each packet. The frames, when pieced together, will form an actual image file, sound file, etc.

There's a lot to do first, but I wanted to share that idea with you for the future. I've been talking about SystemPackets for like, six months and just implemented them a couple weeks ago,😄

PR #77, opened 2022-07-26
Prefix debug messages with millis() timestamp

aviateur17:
If you want this would add a runtime timestamp in front of the debug messages for additional troubleshooting capability.   You can reject this if you don't want this capability.

example:
```
124433:  Connecting to WiFi...
```

### DIFF
--- a/FDRS_Gateway/fdrs_functions.h
+++ b/FDRS_Gateway/fdrs_functions.h
@@ -365,30 +365,29 @@ void getLoRa() {
       }
       receivedLoRaMsg++;
       // Evaluate CRC
-      if(packetCRC == 0xFFFF) { // CRC is set that destination does not want ACK so do not send.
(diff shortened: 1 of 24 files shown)


aviateur17:
Sorry, didn't mean for this PR to be dependent upon the other PRs.  I will try better in the future to make the PRs relative to what is already in dev and not any PRs that have not been approved.

timmbogner:
The Arduino serial monitor has a timestamps option, so this might be a bit superfluous. 

Until I started working with Andreas on the video examples, I was updating the whole repo to GitHub using the "upload file" dialog on chrome... Basically I'm just figuring out Github too, so don't worry about mistakes. It's really overwhelming and I also screw up at least 25% of the time.

aviateur17:
> The Arduino serial monitor has a timestamps option, so this might be a bit superfluous.

Sounds good.  I use PlatformIO with VSCode and that does not have timestamps and others may use PuTTY or some other terminal emulator but no biggie. I'll just keep those timestamps on my side.

PR #84, opened 2022-07-29
Feature add: LoRa SystemPacket ping reply

aviateur17:
Added code to allow LoRa sensor send ping to LoRa gateway and receive ping reply from LoRa gateway back to the LoRa sensor via SystemPacket cmd_ping.

For the cmd_ping SystemPacket I utilized param as follows:
param = 0 is for ping request (Sensor -> Gateway for now)
param = 1 is for ping reply (Gateway -> Sensor for now)

gateway output:
```
Incoming LoRa. Size: 11 Bytes, RSSI: -45dBm, SNR: 9.00dB, PacketCRC: 0x5a77, Total LoRa received: 16, CRC Ok Pct 100.00%
We have received a ping request from 0xee42, Replying.
Transmitting LoRa message of size 11 bytes with CRC 0x6d6a to LoRa MAC 0xee42
Incoming LoRa. Size: 20 Bytes, RSSI: -44dBm, SNR: 9.50dB, PacketCRC: 0xa5d3, Total LoRa received: 17, CRC Ok Pct 100.00%
Sensor address 0xee42(hex) does not want ACK
Sending MQTT.
```

sensor output:
```
Transmitting LoRa message of size 11 bytes with CRC 0x5a77 to destination 0xee05
LoRa ping sent to gateway address: 0xee05
Incoming LoRa. Size: 11 Bytes, RSSI: -45dBm, SNR: 9.50dB, PacketCRC: 0x6d6a
We have received a ping reply via LoRa from address 0xee05
LoRa Ping Returned: 56ms.
Id: 42 - Type: 3 - Data loaded: 21.20
Id: 42 - Type: 1 - Data loaded: 42.40
Sending FDRS Packet!
Transmitting LoRa message of size 20 bytes with CRC 0xa5d3 to gateway 0xee05
 LoRa sent.
Sleepytime!
 Delaying.
```

fdrs_sensor.h is now 500 lines.
fdrs_functions.h is now 1055 lines.
Do we want to start thinking about splitting those up into different files?  Maybe LoRa stuff in it's own file for each the sensor and the gateway?   As we add capabilities the two files above are starting to look more and more similar.

### DIFF
--- a/FDRS_Gateway/fdrs_functions.h
+++ b/FDRS_Gateway/fdrs_functions.h
@@ -20,13 +20,11 @@ enum {
   event_lora2
 };
 
-
 enum crcResult{
   CRC_NULL,
   CRC_OK,
   CRC_BAD,
 } returnCRC = CRC_NULL;
-
 
 enum {
   cmd_clear,
@@ -39,13 +37,13 @@ enum {
 #define DBG(a) (Serial.println(a))
 #else
 #define DBG(a)
-#endif // FDRS_DEBUG
+#endif
 
 #if defined (ESP32)
 #define UART_IF Serial1
 #else
 #define UART_IF Serial
-#endif // ESP32
+#endif
 
 // enable to get detailed info from where single configuration macros have been taken
 #define DEBUG_NODE_CONFIG
@@ -204,6 +202,7 @@ DataReading theData[256];
 uint8_t ln;
 uint8_t newData = event_clear;
 uint8_t newCmd = cmd_clear;
+bool is_ping = false;
 
 #ifdef USE_ESPNOW
 DataReading ESPNOW1buffer[256];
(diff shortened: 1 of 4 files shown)


timmbogner:
>Do we want to start thinking about splitting those up into different files?
Absolutely. At the moment I have some work done on a two-way communication branch, but when that is finished or at a stopping point, let's plan on talking about and finalizing the division and organization of the files. I think that a file for each interface (serial, LoRa, MQTT, SD, will be much simpler, and easier to expand.

thefeiter:
It would also be useful to somehow find out, if two gateways are in reach for each other. So you can see, if the chain of gateways to a remote location will work.

aviateur17:
> It would also be useful to somehow find out, if two gateways are in reach for each other. So you can see, if the chain of gateways to a remote location will work.

Yes, it is a good idea but I think this starts to approach a mesh network situation and I don't think that's the intent of this project.

FDRS_Gateway/fdrs_functions.h
```diff
@@ -527,8 +574,33 @@ void transmitLoRa(uint16_t* destMac, DataReading * packet, uint8_t len) {
   LoRa.write((uint8_t*)&pkt, sizeof(pkt));
   LoRa.endPacket();
   }
+#endif
+
+#ifdef USE_LORA
```
aviateur17, line 579: @timmbogner,  here is the commit to add the issue function to fdrs_functions.h

PR #86, opened 2022-07-30
Dev - sketches in examples folder of current dev branch do not compile.

Gulpman:
Example sketches do not compile. 

I could fix some of the issues but am stuck with what is left.

Here's the compile output from Arduino IDE for this pull request:

```
In file included from D:\_sync\Sascha\OneDrive\Documents\Arduino\libraries\Farm-Data-Relay-System\examples\1_LoRa_Sensor\1_LoRa_Sensor.ino:10:0:
D:\_sync\Sascha\OneDrive\Documents\Arduino\libraries\Farm-Data-Relay-System/fdrs_sensor.h: In function 'crcResult getLoRa()':
D:\_sync\Sascha\OneDrive\Documents\Arduino\libraries\Farm-Data-Relay-System/fdrs_sensor.h:199:33: error: 'SystemPacket' was not declared in this scope
   if ((packetSize - 6) % sizeof(SystemPacket) == 0 && packetSize > 0) {  // packet size should be 6 bytes plus multiple of size of SystemPacket
                                 ^
D:\_sync\Sascha\OneDrive\Documents\Arduino\libraries\Farm-Data-Relay-System/fdrs_sensor.h:206:18: error: expected ';' before 'receiveData'
     SystemPacket receiveData[ln];
                  ^
D:\_sync\Sascha\OneDrive\Documents\Arduino\libraries\Farm-Data-Relay-System/fdrs_sensor.h:216:14: error: 'receiveData' was not declared in this scope
       memcpy(receiveData, &packet[4], packetSize - 6);   //Split off data portion of packet (N bytes)
              ^
In file included from D:\_sync\Sascha\OneDrive\Documents\Arduino\libraries\Farm-Data-Relay-System\examples\1_LoRa_Sensor\1_LoRa_Sensor.ino:10:0:
D:\_sync\Sascha\OneDrive\Documents\Arduino\libraries\Farm-Data-Relay-System/fdrs_sensor.h: In function 'void sendFDRS()':
D:\_sync\Sascha\OneDrive\Documents\Arduino\libraries\Farm-Data-Relay-System/fdrs_sensor.h:340:49: error: invalid conversion from 'uint16_t {aka short unsigned int}' to 'uint16_t* {aka short unsigned int*}' [-fpermissive]
   transmitLoRa(gtwyAddress, fdrsData, data_count);
                                                 ^
[...]

PR #88, opened 2022-07-31
Fix Arduino compile issue as mentioned in issue #86

aviateur17:
Synchronization between two fdrs_sensor.h files which caused issues in Arduino compilation.   Also not sure why crc16_update prototype on line 105 is required by Arduino.  I just added it instead of figuring out why Arduino needs it.   PlatformIO does require function prototypes if the functions are used before they are defined but my understanding was that Arduino handled that situation without requiring the prototypes.


aviateur17:
@Gulpman, I tried one or two of the gateway sketches and thought they did compile.  I will go and check them all right now.  Sorry for missing this.

timmbogner:
Just let me know when things are ready to merge.

Many thanks!

aviateur17:
@Gulpman, I tested examples 1 through 5 and the first example under Sensor_Examples and they compile correctly with the function prototypes added in.   It seems that the Universal Gateway stuff is being maintained by someone else so I didn't try to compile those. I made a note of why the Stress test would not compile in #86. The sketches in the Controllers folder do not seem like they are being maintained so I'm not sure if those are to be tested or not.   This is one of the reasons why the project structure is a bit confusing.

PR #89, opened 2022-08-02
Never mind... I digress

timmbogner:
This should be the last time we deal with this, as we're going to remove all redundant files immediately! 😅

This PR seemed like an easy way to compare the two... I dunno. We'll see how it goes.

### DIFF
--- a/fdrs_functions.h
+++ b/fdrs_functions.h
@@ -192,16 +192,18 @@ unsigned long ackOkLoRaMsg = 0;     // Number of total LoRa packets with valid C
 #endif
 
 #if defined (USE_SD_LOG) || defined (USE_FS_LOG)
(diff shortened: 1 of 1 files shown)


timmbogner:
Never mind, I thought there were more discrepancies. I think the file from the Gateway folder was actually just newer and this can indeed be merged after all.  Correct me if I'm wrong @thefeiter

timmbogner:
Here's a peek at the FDRS hardware division. I designed these PETG enclosures over the winter and splurged on colors to make them look like 90's tech😎

Yes, FDRS has a logo! ☺️ It never printed well, though. 

IMG_20220801_215945941
IMG_20220801_220121509
IMG_20220801_220024182

PR #91, opened 2022-08-04
decouple READING_ID from LoRa address

aviateur17:
Change READING_ID back to integer and decouple it from the LoRa address.

### DIFF
--- a/FDRS_Sensor/README.md
+++ b/FDRS_Sensor/README.md
@@ -14,8 +14,8 @@ If available and enabled, the device enters deep-sleep. If ```#DEEP_SLEEP``` is 
 
 ## Options
 
-### ```#define READING_ID  0xnn```
-The identifier of this individual device. Should be a 8-bit hexidecimal value (0x00 - 0xFF).
+### ```#define READING_ID  n```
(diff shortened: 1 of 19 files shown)


timmbogner:
oop, missed this while I was yapping over in the other topic 😶

Looks good for now.  We do need to make them unique, though. 

One quick confession: I thought there was a way to tell the compiler to insert a random number into the sketch, like ``` __RANDOM__ ``` or something. I just looked and this does not exist. 

 ``` __DATE__``` and ```__TIME__```  do exist, however and I think we can process those into a unique address.

timmbogner:
After some thought and research, I think the best route to take is to borrow the last two digits of the MAC address if it is an ESP. If it is not, I found the (ArduinoUniqueID library)[https://github.com/ricaun/ArduinoUniqueID], which should get us unique addresses for other chipsets.

As a note, ArduinoUniqueID was designed for a library named LoRaNow, which is much like what we're doing. I was made aware of LoRaNow a while back, but it lacks a "sender address" in the packet structure, making it unsuitable for FDRS. Definitely worth looking at, though.

PR #100, opened 2022-08-26
Minor fix of logical && condition test

Gulpman:
Logical and check for preprocessor directives was not done the right way. Fixed that.


timmbogner:
I'm a big fan of this functionality, thanks! Definitely agree that it should be enabled by default.

Nodes will soon have several parameters set/changed after boot, so I hope in the future we can work this in as a status readout. For example: While working with a controller, the user may want to know which reading_ids it is currently subscribed to, or whether it has successfully registered with a gateway.

PR #122, opened 2022-11-29
Conflict with pin12

lamasseriadipolverara:
change the  pin 12 to pin 16, because pin 12 affect the boot of the device (esp32 wroom - 32e)


timmbogner:
Thanks for pointing that out! However, pin 16 would cause a conflict on ESP8266. The pins are laid out for the WeMos D1 Mini that I use for testing.

 I changed it to pins that are better but still not good for ESP32. According to these docs, you're going to be dealing with PWM signals on pins 5 and 14 at boot: [32 ](https://randomnerdtutorials.com/esp32-pinout-reference-gpios/)[8266](https://randomnerdtutorials.com/esp8266-pinout-reference-gpios/),

PR #139, opened 2023-02-26
2023 01 lora improvements

aviateur17:
Added PingFDRS to LoRA node code so that Node can ping gateway
Moved ESP-NOW pingFDRS code to fdrs_node_espnow since the code was split from one header file to several header files
Tweaked sleepFDRS function so that it doesn't say sleepytime since if DEEP_SLEEP is commented out the controller does not actually go to sleep.
Added crcResult return type to transmitLoRa functions so that the calling functions can actually know whether the packet was transmitted successfully and that information can be used in calling functions.   This seems to be moreso used in nodes and not so much in gateways at the current time but gateway use of crcResult return type can be developed at a later time if need be.
Used Arduino 1.8.19 to compile gateway, sensor, and controller sketches on ESP32 board type with ESP-NOW enabled and separately with LoRa enabled. 
Tweaked ESP-NOW debug message to properly include units for ping time and MAC address for ESP-NOW address being pinged.

### DIFF
--- a/src/fdrs_gateway.h
+++ b/src/fdrs_gateway.h
@@ -41,7 +41,6 @@ DataReading theData[256];
 uint8_t ln;
 uint8_t newData = event_clear;
 uint8_t newCmd = cmd_clear;
-bool is_ping = false;
 
 DataReading fdrsData[256]; // buffer for loadFDRS()
 uint8_t data_count = 0;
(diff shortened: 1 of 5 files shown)


timmbogner:
Coincidence: Right as you were posting this I was finishing up OLED display for nodes and noticing that "Sleepytime!" was now sort of wasting screen space.

I'll look at this and pull it tomorrow morning.

PR #142, opened 2023-02-27
Added FDRS Timekeeping

aviateur17:
You can reject if you want to add this to a different branch in order to wait to include in the main release branch.   This adds code in Gateway for NTP Time server queries via WiFi.   Compiled in Arduino 1.8.19 with and without USE_WIFI defined.


timmbogner:
I love that you did it all without the time.h library. That definitely is better than I could have made it. My (literally) only concern presently is the DNS config. Is there any way to do that automatically? We'll at least have to make the definition a string probably, to be beginner-friendly and match MQTT config.
That's really my only feedback. Otherwise it's great. I'll probably just leave this open through the new release, and we can discuss further details like how to share the time in the discussion thread.

aviateur17:
> I love that you did it all without the time.h library. 

Yeah, I wanted to write the code with as little dependencies as possible to simply the compile process and size of the code

>My (literally) only concern presently is the DNS config. Is there any way to do that automatically? We'll at least have to make the definition a string probably, to be beginner-friendly and match MQTT config. That's really my only feedback. Otherwise it's great. 

We can default the DNS server to a public DNS server like Google or some other DNS server if you'd like.  That way we minimize the chance that people would have issues but still let them set the DNS server if they'd like.   If it's the format of the IP address with the commas I'll have to go back and look at how I can change it into a string.  It's most likely possible if that is desired.

>I'll probably just leave this open through the new release, and we can discuss further details like how to share the time in the discussion thread.

Sounds good!

aviateur17:
>My (literally) only concern presently is the DNS config. Is there any way to do that automatically? We'll at least have to make the definition a string probably, to be beginner-friendly and match MQTT config. That's really my only feedback.

Fixed in commit https://github.com/timmbogner/Farm-Data-Relay-System/pull/142/commits/cce5fe57d4142d07621341645a2fa6789fba22a5

aviateur17:
In my commit https://github.com/timmbogner/Farm-Data-Relay-System/pull/142/commits/dcda381ea743498fb69474efe74b21505c68aee1 in the gateway_time.h and node_time.h files there is a couple of loadFDRS() calls with the current time and slew rate.  I've disabled those by default as I think casual users would not care about that information.  But just FYI if you want to do something with those like only enable when DEBUG is enabled and disable otherwise?

My thoughts on next feature adds in this topic:
1. share time via ESP-NOW from gateway to nodes
2. Option to disable DST for areas that do not observe DST
3. Add timestamps to DataReadings while maintaining backwards compatibility for DataReadings that do not have timestamps

I'm not sure how soon I'll get to those features. I have a couple of others I want to work with such as enable OTA updates for WiFi enabled gateways and perhaps BLE UART communication.  I'll try to add the ESP-NOW part in the next few days and then the others may be a week or two or so.

aviateur17:
@timmbogner, question on events/actions.   Based on the repeater examples it seems that the convention is that ESPNOW1_ACT and LORA1_ACT are directed towards the nodes and, oppositely, ESPNOW2_ACT and LORA2_ACT are directed towards the front end?   I'm thinking in context of periodically sending the time from the front-end to the nodes that the actions to be taken should be under ESPNOW1_ACT and LORA1_ACT but the documentation doesn't really specify any directionality but the examples seem to imply what I stated above.  I think up to now we really haven't had much or any traffic in the front-end to node direction so it hasn't been important to clarify that.   So I'm thinking that I'm going to use ESPNOW1_ACT and LORA1_ACT as events/actions to determine where to direct the time data.  Let me know your thoughts please.  Thanks!

timmbogner:
Great question. It's actually the **_big_** question for me in terms of timekeeping. 
First, you're right. The convention I've been using is for LORA1 to be towards the front-end and LORA2 towards the nodes and repeaters. But I would like to keep it a loose convention, never forcing the user to set it up any given way.  Eventually I think I might add more neighbors, in which case having a pre-determined role for any of them a little unclear to a new user.

So I reckon there are two paths to think down here... 
1. We re-use the existing routing configuration for SystemPackets (or specifically just time info). Essentially building a SystemPacket alternative into sendSerial(), sendESPNowNbr(), etc. I've thought about this, but it's not my first choice.
2. We more or less flood all interfaces with the time when we get it, sending the time out to UART, ESP-NOW/LoRa neighbors, broadcast to LoRa nodes, send to ESP-NOW peers, just goin nuts. The exception being the interface in which the time was received. I _think_ this would avoid loops. This would also enable a device with an RTC located remotely to send a valid time inward to nodes that are closer to the front-end.

Let me know what you think!

aviateur17:
I totally agree with you that we should keep it as flexible as possible for any future expansion and  your  item 2 above is the way to go.  I think for now I'll set up ESP-NOW to go to Neighbor 2 (acting as repeater) **and** peers (acting as gateway).  For LoRA I will send to Neighbor 2 (acting as repeater) **and** broadcast (acting as gateway).   If, in the future,  you tweak how things work it can be easily changed as I will try to neatly put the code in functions.   

On a slightly different topic, as the code matures I think we may need to revisit function names to match what LoRa is doing to what ESP-NOW is doing as closely as possible for clarity.   We have functions for initialization, functions for looping, lower level functions that handle the nitty gritty data encapsulation details and higher level functions that call the lower level functions.  as more functionality is added it might be good to standardize on the names.   Something that came to my mind while reviewing the code and something to ponder for the future.

timmbogner:
I think it should send to all neighbors, just in case. 
A fringe configuration that is currently possible is that the gateway connected to the internet has two different neighbors, each being a repeater going out in different directions. This is feasible with WiFi-to-LoRa gateways and Ethernet-to-ESP-NOW gateways.

We may need to better define a condition where a neighbor is not present in configs, which would actually be good. ```#define LORA_NEIGHBOR_1    0x00``` is what I define a "null" neighbor as, since in current usage 0x00 should only be an MQTT gateway and thus not listening to any radios. That's not great though, and it would be better to be able to just disable it. I don't think a 'NULL' or '-1' value would be safe, but I haven't looked too far into it.

timmbogner:
>On a slightly different topic, as the code matures I think we may need to revisit function names to match what LoRa is doing to what ESP-NOW is doing as closely as possible for clarity. We have functions for initialization, functions for looping, lower level functions that handle the nitty gritty data encapsulation details and higher level functions that call the lower level functions. as more functionality is added it might be good to standardize on the names. Something that came to my mind while reviewing the code and something to ponder for the future.

I'll definitely ponder this. I noticed when making functions for the recent reshuffle that I had begun ```setting_names_like_this()``` instead of ```likeThis()``` as I had usually done starting out. There are certain conventions like '-FDRS' as a suffix (```beginFDRS()```) is used for "public" functions. I'm always trying to prepare to shift the library to being class-based eventually, and that's my attempt to emulate what might become for example: ```FDRS.begin();```

aviateur17:
> I think it should send to all neighbors, just in case. A fringe configuration that is currently possible is that the gateway connected to the internet has two different neighbors, each being a repeater going out in different directions. This is feasible with WiFi-to-LoRa gateways and Ethernet-to-ESP-NOW gateways.

I'm just afraid that if we send data back in the other direction, towards the front end, that will cause a loop.   In general, all of the data should be sent in one direction, but there is no directionality in the code, because up to now there was no reason to send information from the front end down to the node.

timmbogner:
>I'm just afraid that if we send data back in the other direction, towards the front end, that will cause a loop. 

I'll admit, that possibility is why I'm not 100% sure of myself. Preventing the time data from being returned to its source is my means of preventing this. So if the time was received from ESPNOW1, it would never be sent back to that device, only ESPNOW2 and peers.

timmbogner:
I should say though, you're welcome to keep it how you think is best for the time-being.

aviateur17:
Going through the code to implement timekeeping via ESP-NOW seems like there is quite a bit to be modified.  I wonder if this would be better as a different PR.  The issue is that this PR was started before the last major commit to main branch so there are things that were committed to main that are conflicting with this code.  So I think the next related commit to this might be better committed against main and wait till this PR is also completed.

timmbogner:
Dangit... holdupasec

timmbogner:
Soo... I thought that first commit had resolved the conflicts, then it added the other when I pushed it. My damage control didn't help. I dunno 😣

[Here is a branch](https://github.com/timmbogner/Farm-Data-Relay-System/tree/timekeeping-fixed) without the extra stuff that I think should be okay? Sorry... dunno what to do at this point. I honestly didn't think I was pushing it directly to your fork, I sorta thought there was an intermediate branch for this pull request.

aviateur17:
No problem. I think I can reject the pr and push again against the new main
branch. I'll look at it probably tomorrow.

aviateur17:
Okay, I think I fixed the conflicts but I need to go through and make sure stuff compiles and then add my latest round of ESP-NOW changes.   So don't merge this in yet.  In the future, if I see conflicts with my PRs I know how to fix them :)

timmbogner:
> In the future, if I see conflicts with my PRs I know how to fix them :)
:D I was so close to leaving it alone, but was convinced I'd figured out how to do it.

> Guess I can just update the "base branch" and we're good to go.
Oh wow... so simple!

aviateur17:
Not done yet.  Still need to correct a DST/Std time issue and do more testing with ESP-NOW and LoRa.  It might be a few days until I come back to this.

aviateur17:
@timmbogner, I was hoping to implement all of this without adding any libraries but I'm not sure how to "easily" calculate whether the current date is within daylight savings or not.  I was hoping the tm struct element tm_isdst would work but I'm not sure how it knows what time zone is being used.  Using ESP libraries aren't going to work because I think this is more of an Arduino centric project so I would propose using the "de facto" Arduino timezone library https://github.com/JChristensen/Timezone   

Any comments on that or other ideas?

timmbogner:
I am still getting caught up... There is one commit I need to make to fix custom SPI, then I plan to get up to speed on this PR and the others. 

I really appreciate you taking this on though, because I clearly wouldn't have had time to start this timekeeping project for a good while.

timmbogner:
I see the conundrum about DST.

>I was hoping the tm struct element tm_isdst would work but I'm not sure how it knows what time zone is being used.

My loose understanding is that if tm_isdst is 1, then *if* you are observing daylight saving time, you would want to set the time ahead an hour. So we could have a ```CONFIG_OFFSET_HOURS -6``` , then a config ```USE_DST```. if tm_isdst is 1 and USE_DST is defined, then the time gets moved ahead.

I could be wrong about that.

I went over this once and will do so a couple more times to get a good feel for what you're up to. It all looks cool! The only concern I have so far is that[ sendESPNow() was documented and actually was used commonly in old configurations](https://github.com/timmbogner/Farm-Data-Relay-System/blob/main/extras/Gateway.md#sendespnow0xnn). I'm leaning towards the more portable neighbor functions now, but I still think we'll need to make sure that ```void sendESPNow(uint8_t address)``` will be available for backwards compatibility.

timmbogner:
I just remembered that sendESPNow actually was sendESPNOW for most of its existence, so maybe you can reuse it. It could change to sendESPNowDirect() maybe...

aviateur17:
> I am still getting caught up... There is one commit I need to make to fix custom SPI, then I plan to get up to speed on this PR and the others.

Now I'm having issues with the ESP32 hanging while loading LoRa, I think it's related to SPI interface as when I remove LoRa it boots just fine.  I'm trying to figure that out. The main branch works just fine so it's some change that has been made since main.

aviateur17:
> My loose understanding is that if tm_isdst is 1, then _if_ you are observing daylight saving time, you would want to set the time ahead an hour. So we could have a `CONFIG_OFFSET_HOURS -6` , then a config `USE_DST`. if tm_isdst is 1 and USE_DST is defined, then the time gets moved ahead.

The problem is that I don't see a way to set a time zone such that tm_isdst knows when to change from DST to STD time.  Ideally we shouldn't have to download the controller to make this transition.  Without a ton of coding I'm not sure how to do this so I'd look at adding the library mentioned earlier.

> I went over this once and will do so a couple more times to get a good feel for what you're up to. It all looks cool! The only concern I have so far is that[ sendESPNow() was documented and actually was used commonly in old configurations](https://github.com/timmbogner/Farm-Data-Relay-System/blob/main/extras/Gateway.md#sendespnow0xnn). I'm leaning towards the more portable neighbor functions now, but I still think we'll need to make sure that `void sendESPNow(uint8_t address)` will be available for backwards compatibility.

I see sendESPNow as a "lower level function" that should not be called directly  but only referred to by other "higher level" functions.   It basically encapsulates the data type handling needed by higher level functions and handles calls to the Espressif API calls.   I could be missing something though, where would changing the signature  or parameters of the function affect the backward compatibility?

timmbogner:
>Now I'm having issues with the ESP32 hanging while loading LoRa, I think it's related to SPI interface as when I remove LoRa it boots just fine. I'm trying to figure that out. The main branch works just fine so it's some change that has been made since main.

I've made a couple SPI fixes recently so check main's recent history for clues.


> such that tm_isdst knows when to change from DST to STD time.

My understanding is that tm_isdst will be known and set by the time server, essentially it should always be 1 during the summer. The boolean could be translated like "are we within the time period when clocks are generally set forward? 1 if yes, 0 if no". It doesn't care where we are. I'm assuming the "spring forward/fall back" date is known by the server and it will change the value on that date. I'm curious though, if this is indeed the case, how RTCs know. Would they have it pre-coded? 
To be fair : the tutorial I learned that from was for Python, and the accompanying example didn't make sense to me. I could be totally off base.

> So I guess what I'll have to do is keep that existing one for backwards compatibility.

Admittedly, I forgot that I didn't really care when I changed it from sendESPNOW to sendESPNow, so it doesn't _really_ matter now.
What if you changed the low-level version to transmitESPNow? That would have the benefit of matching LoRa a little bit.

aviateur17:
> My understanding is that tm_isdst will be known and set by the time server, essentially it should always be 1 during the summer. The boolean could be translated like "are we within the time period when clocks are generally set forward? 1 if yes, 0 if no". It doesn't care where we are. I'm assuming the "spring forward/fall back" date is known by the server and it will change the value on that date. I'm curious though, if this is indeed the case, how RTCs know. Would they have it pre-coded? To be fair : the tutorial I learned that from was for Python, and the accompanying example didn't make sense to me. I could be totally off base.

All the time server sends is the number of seconds since 1900 in UTC time.  There's no timezone or DST information so it's up to the application to determine all of that.
 
> Admittedly, I forgot that I didn't really care when I changed it from sendESPNOW to sendESPNow, so it doesn't _really_ matter now. What if you changed the low-level version to transmitESPNow? That would have the benefit of matching LoRa a little bit.

Good thought on the rename.  I had a similar thought.

aviateur17:
Looks like DST information isn't in the NTP standard UDP packet data:
https://www.meinbergglobal.com/english/info/ntp-packet.htm

timmbogner:
> Looks like DST information isn't in the NTP standard UDP packet data

Yeah makes sense now that I think about it. I was forgetting how it all worked 🙄
Most of what I said here was pretty wrong as it turns out, so I'ma go back to researching. If we need the library then so be it, but I'd rather avoid dependencies of course.

timmbogner:
Spitballing:
It looks like DST is divided between [two major observation periods](https://en.wikipedia.org/wiki/Daylight_saving_time_by_country).
What if we had ```CONFIG_OFFSET_HOURS ``` and then the options of ```USE_DST_US```  or ```USE_DST_EU```. We can code in the logic for either one's start/end dates (unless this is too laborious) and then be able to figure it all out ourselves. This leaves out Egypt, Paraguay, Chile, NZ, and parts of Australia of course, so a bit messy.

aviateur17:
Yeah, that was going to be my approach.  So use the NTP time to calculate current date and see if it is in DST or STD time.  The problem is the time change is not a specific date but like the second Sunday in March and the first Sunday in November.  So the date of change is different in every year.  So then you'd have to calculate the date for the second Sunday in March in the current year.  So to calculate that you feed what you can into the timeInfo struct.  I'm not sure that it will take in a day of week and a month and a year and give you a specific date.  I think that is where I got stuck.

timmbogner:
>problem is the time change is not a specific date

Yeah that's the step I thought could be too laborious. At the least you could just use the Timezone library for now and hope to come up with something else later. Not the end of the world if we need to use the library.

aviateur17:
Could plug in a date like March 10th and then get the day of week.  Then add or subtract the number of days from the current day of week to get to Sunday to get the date needed.  Then calculate the difference between today's date and the known date to see if we are in DST or not.  I'll see if I can get that to work.

aviateur17:
US DST testing output:
US DST Start - 2nd Sunday in March - 02:00 local time
US DST End - 1st Sunday in November - 02:00 local time

```
The current local date/time is: Wed Mar  1 17:33:39 2023 STD

The current local date/time is: Sun Mar 12 01:59:55 2023 STD
Time change from STD -> DST
The current local date/time is: Sun Mar 12 03:00:05 2023 DST

The current local date/time is: Sun Nov  5 01:59:55 2023 DST
Time change from DST -> STD
The current local date/time is: Sun Nov  5 01:00:05 2023 STD
EndTest
```

EU DST Testing output:
(local time = UTC +2)
EU DST Start - last Sunday in March - 01:00 UTC
EU DST End - last Sunday in October - 01:00 UTC

```
The current UTC date/time is: Wed Mar  1 23:33:39 2023
The current local date/time is: Thu Mar  2 01:33:39 2023 STD

The current UTC date/time is: Sun Mar 26 00:59:54 2023
The current local date/time is: Sun Mar 26 02:59:55 2023 STD
Time change from STD -> DST
The current UTC date/time is: Sun Mar 26 01:00:04 2023
The current local date/time is: Sun Mar 26 04:00:05 2023 DST

The current UTC date/time is: Sun Oct 29 00:59:55 2023
The current local date/time is: Sun Oct 29 03:59:55 2023 DST
Time change from DST -> STD
The current UTC date/time is: Sun Oct 29 01:00:05 2023
The current local date/time is: Sun Oct 29 03:00:05 2023 STD
EndTest
```

aviateur17:
Was having issues with MQTT messages not sending - just generating the error.  Figured out that the Arduino PubSubClient default size is 256 bytes and need to increase the default buffer by using setBufferSize.

https://pubsubclient.knolleary.net/api#setBufferSize

timmbogner:
> Was having issues with MQTT messages not sending - just generating the error. Figured out that the Arduino PubSubClient default size is 256 bytes and need to increase the default buffer by using setBufferSize.

Interesting! Does that mean, essentially, that sending >256 bytes of JSON wasn't ever working? Yikes 😬
I thought I had sent big packets via MQTT before, but thinking about it... not sure when. I'm not able to test it presently, but I'll look at that ASAP.

aviateur17:
I think everything is working now so I'm going to get the last changes into the commit and then make sure everything compiles in Arduino and then should be done on this.

aviateur17:
Of course my return type esp_err_t isn't valid for ESP8266 and the return types between ESP32 and ESP8266 for some functions are not consistent so I think I'm going to have to make another custom type.

timmbogner:
Hm... that sounds familiar now that you mention it. One of several mysterious 32 vs 8266 incongruencies.
Can you copy the declaration (or whatever) from ESP32 into _fdrs_datatypes.h_ and wrap it with ```#ifdef ESP8266```?

I'm planning to devote some upcoming rainy days to testing all of this new stuff out, including the other PRs.

aviateur17:
> Hm... that sounds familiar now that you mention it. One of several mysterious 32 vs 8266 incongruencies. Can you copy the declaration (or whatever) from ESP32 into _fdrs_datatypes.h_ and wrap it with `#ifdef ESP8266`?

Very good idea, I was thinking exactly the same.  I'm going to be on vacation for most of this week so I won't be able to finish this PR until late this week at the earliest.  I think it's all complete except to make sure everything compiles properly in Arduino.

timmbogner:
I just jumped into testing this and am working towards ethernet compatibility. At first,  ``` if(WiFi.status() == WL_CONNECTED) { ```  never became true [here](https://github.com/timmbogner/Farm-Data-Relay-System/blob/f93f41b8aa474715e95214e21266bbee7c3a04b1/src/fdrs_gateway_wifi.h#L156). I added ``` || eth_connected``` and it made it to the next step, but times out getting NTP response.  Any ideas?

timmbogner:
Scratch that! I just looked over and it was displaying the time. Must have been a one-off error.

aviateur17:
Timm, while compiling this branch for Adafruit Feather M0 (SAMD21 micro) it seems that the settimeofday() function does not work for that microcontroller.  The compilation succeeds but the linking step at the end fails.  I'm not sure without that function if the time will be set in the SAMD21 micro.   Do I just put some #ifdefs around that function call and we can have someone test the functionality on the SAMD21?   I know you would like to be as microcontroller agnostic as possible.   It seems like there would be another library needed to use the RTC functionality in the SAMD21 to keep time in that micro.

timmbogner:
I'm not sure yet. This gave me the idea to test compilation for Pico, and predictably "esp_err_t" isn't defined for it (for starters).
I think that it would be okay to use #ifdefs to switch between chip-specific RTC routines. This might come up again when adding support for varying external RTC modules.
 If there is a library needed, it would be okay to require it for timekeeping, but my hangup is having it required to install for users who may not need timekeeping.

Looks like AVR doesn't have "sys/time.h" either. 

For now, should we stipulate that timekeeping is only for Espressif chips? Have you confirmed it on anything else?

I have been playing with the _gateway_ timekeeping exclusively so far and everything seems fine on 8266, 32, and S3. My only thought right now:

I have an MQTT gateway with an ESP-NOW repeater. The MQTT gateway gets the accurate time and gives it to the repeater. However, the repeater will then end up giving its skewed time back to the MQTT gateway. What if instead of just sharing the time periodically, it was restricted to only happening when the gateway is given a verified-accurate time? So devices would only share the time when they receive a "fresh" time, either by NTP or a gateway sending one.  This applies for sharing amongst gateways... nodes would be different and I haven't gotten to them.

aviateur17:
> ... predictably "esp_err_t" isn't defined for it (for starters). 

I think I have the fix for that.  I found that as well and instead of using another struct for esp_err_t I will change that to a typdef as below and that will work for other micros.
```
#ifndef ESP32
typedef int esp_err_t;
#define ESP_FAIL 0
#define ESP_OK 1
#endif
```

>I think that it would be okay to use #ifdefs to switch between chip-specific RTC routines. This might come up again when adding support for varying external RTC modules. If there is a library needed, it would be okay to require it for timekeeping, but my hangup is having it required to install for users who may not need timekeeping.
> 
> Looks like AVR doesn't have "sys/time.h" either.
> 
> For now, should we stipulate that timekeeping is only for Espressif chips? Have you confirmed it on anything else?

I've only confirmed on ESP32 and ESP8266.  I have a Sparkfun ProRF LoRa which has SAMD21 on it but I haven't yet put FDRS on that one.  I think we should say that it's only been tested on Espressif chips and would welcome other feedback.

> I have an MQTT gateway with an ESP-NOW repeater. The MQTT gateway gets the accurate time and gives it to the repeater. However, the repeater will then end up giving its skewed time back to the MQTT gateway. 

Yeah, right now the time is sent via ESP-NOW back to both neighbors and all registered nodes.  We can change that if we figure out a better method.

> What if instead of just sharing the time periodically, it was restricted to only happening when the gateway is given a verified-accurate time? So devices would only share the time when they receive a "fresh" time, either by NTP or a gateway sending one. This applies for sharing amongst gateways... nodes would be different and I haven't gotten to them.

Yes, we can only have the time sent when it is refreshed via NTP.  That won't currently prevent it being sent from the ESP-NOW repeater back to the MQTT gateway in your example above though.  I could implement that if the 
```
#define TIME_SEND_INTERVAL 5    // Time, in minutes, between sending out time to remote devices
```
Is set to 0 then only send the time when received by NTP and we can have the default value at 0.  However with introduction of RTC without NTP we will probably need the periodic send.  The other thing is that if, for some reason, the gateways miss the packet being transmitted at the time they won't get the new time for potentially quite awhile.

aviateur17:
> I have an MQTT gateway with an ESP-NOW repeater. The MQTT gateway gets the accurate time and gives it to the repeater. However, the repeater will then end up giving its skewed time back to the MQTT gateway. 

I was wondering if we should classify one of the neighbors as the "upstream" neighbor and one the "downstream" neighbor so that the time is only sent to the downstream neighbor and not back to the upstream neighbor.  I think you said earlier that for now we should keep the setup as generic/flexible as possible.

timmbogner:
>I was wondering if we should classify one of the neighbors as the "upstream" neighbor and one the "downstream" neighbor so that the time is only sent to the downstream neighbor and not back to the upstream neighbor.

I've had a response draft I've been mulling over since yesterday that I'll post now before I head to work. There's actually another option I can get into later that you've brought up involving specific time-routing options.

Here is the scenario(s) I've been thinking about, let me know your thoughts. Am I overlooking some looping possibilities again?

#
Gateway 0x01 boots and has WiFi enabled, so it checks to see if it can reach the time server every x minutes. If it gets a time successfully, it then sends it out to neighbors.
Neighboring Gateway 0x02 gets that time, sets its own clock, and sends it on to its neighbors BUT NOT the one that it received it from (0x01). We actually may want do something similar with nodes, to allow for a situation where a node with a powered RTC feeds the time to the whole system. When a gateway re-sends the time that came from a node, it would need to temporarily remove that node from the ESP-NOW peer list when it shares the time with everyone else. 

So essentially the device that knows that it has the correct time, be it by NTP or RTC, initiates a chain reaction that updates all the devices but never goes backwards.

OR

We keep all the time scheduling stuff, but each gateway now keeps track of where the time came from and sends it everywhere else, but never back to where it came from.

timmbogner:
Other ideas:
The gateway with the known time is the only one to initiate time updates, and it sends to everyone BUT ignores subsequent time updates for a certain period of time. So like, a gateway sends out the time, but then can't do it again for a minute, even if a time update comes in.

The idea I alluded to this morning:
Basically assign where the time goes when it comes in, or when it is scheduled. It would essentially be an "outbound list".

aviateur17:
I like the idea of the gateways/repeaters keep track of the source address of where the time packet was received and when sending it out to other notes checking to make sure that it does not send the time to that address.

timmbogner:
I'm back to it, and first let me mention this is suuuper nifty and fun! Timekeeping is going to be such a powerful feature...

Currently I can't get any device to receive a time via LoRa. Pings work fine, but when the gateway sends a LoRa time packet, nothing happens on the next gateway or any nodes.
My setup is an Ethernet gateway connected to a LoRa32 gateway via ESP-NOW, then to another LoRa32 via LoRa. I have tried with both a gateway and a node for the next device. Timekeeping works a charm up to the LoRa hop.

Any ideas?

aviateur17:
I think there was a hook in LoRa that I forgot to implement.  I think I have it almost ready to go but haven't submitted it yet.  I'm trying to think whether it was in the gateway code, I believe.  So gateways wouldn't be able to send to other gateways without the hook.  I'll try to get that submitted tonight.   Sorry about that.

timmbogner:
No worries! I have been working to get my T-Watch going on the system. One thing I've changed for this is to have the gateway send the time to a controller immediately after registering. I'll push that commit later if it's alright with you, as well as adding the following sketch to the controller examples. It works, but in UTC. Can you point me in the right direction?

[Here ](https://github.com/Xinyuan-LilyGO/TTGO_TWatch_Library/blob/master/src/drive/rtc/pcf8563.cpp)is the RTC library being used, I think ```setDateTime(uint16_t year,
                                uint8_t month,
                                uint8_t day,
                                uint8_t hour,
                                uint8_t minute,
                                uint8_t second)``` may be the useful function.


``` cpp
//  FARM DATA RELAY SYSTEM
//
//  LilyGo T-Watch 2020 Example
//
//  Developed by Timm Bogner (timmbogner@gmail.com) in Urbana, Illinois, USA.
//

#define LILYGO_WATCH_2020_V1

#include <LilyGoWatch.h>
#include "fdrs_node_config.h"
#include <fdrs_node.h>

TTGOClass *ttgo;
TFT_eSPI *tft;
PCF8563_Class *rtc;

uint32_t interval = 0;

void fdrs_recv_cb(DataReading theData) {
  DBG("ID: " + String(theData.id));
  DBG("Type: " + String(theData.t));
  DBG("Data: " + String(theData.d));
}

void setup() {
  Serial.begin(115200);
  //Get watch instance
  ttgo = TTGOClass::getWatch();
  // Initialize the hardware
  ttgo->begin();
  // Turn on the backlight
  ttgo->openBL(); 
  // Receive as a local variable for easy writing
  rtc = ttgo->rtc;  
  tft = ttgo->tft;

  beginFDRS();
  pingFDRS(1000);
  addFDRS(fdrs_recv_cb);
  subscribeFDRS(READING_ID);
  delay(100);
  rtc->syncToRtc();
}

void loop() {
  loopFDRS();

  if (millis() - interval > 1000) {
    interval = millis();
    tft->setTextColor(TFT_BLUE, TFT_BLACK);
    tft->drawString(rtc->formatDateTime(PCF_TIMEFORMAT_MM_DD_YYYY), 50, 200, 4);
    tft->setTextColor(TFT_GREEN, TFT_BLACK);
    tft->drawString(rtc->formatDateTime(PCF_TIMEFORMAT_HMS), 5, 75, 7);
  }
}
```

timmbogner:
Sorry, I should have been more specific with my issue. Everything in that sketch works, except the display shows the time in UTC, which is what syncToRtc() sets it as. Is there an easy way to set it to the date/time adjusted for timezone and DST?

aviateur17:
> Sorry, I should have been more specific with my issue. Everything in that sketch works, except the display shows the time in UTC, which is what syncToRtc() sets it as. Is there an easy way to set it to the date/time adjusted for timezone and DST?

The way I've coded the time stuff in FDRS is to store the time in UTC and then convert to local time based on the configured offset and whether the current time is in DST or STD time.  The NTP information is in UTC and I think that's the most flexible way to store the time.  So if you're seeing the time in UTC then it sounds like it's not being converted to your local time.  The printTime() function is what takes the UTC time and converts the output to local time to be displayed.  So I guess the question is whether that library expects local time to be stored in RTC or if it expects UTC.  How does the library handle changeover from DST to STD time and vice versa?  It's almost like it expects the RTC time to be in the local time zone.   So in order to handle that you'd have to get the system time and then offset it by your time zone and then store it to the watch RTC.   I guess what I would probably do is just replace the formatDateTime lines by your own function calls.  So the class calls:

```
snprintf(format, sizeof(format), "%02d-%02d-%02d", t.month, t.day, t.year);
```
and
```
snprintf(format, sizeof(format), "%02d:%02d:%02d", t.hour, t.minute, t.second);
```
so I would substitute those calls with the strftime function call similar to the printTime() function.
Syntax for strftime() can be found at the URL below:
https://www.tutorialspoint.com/c_standard_library/c_function_strftime.htm

```
char strftime_buf[64];
time_t local = time(NULL) + (isDST?dstOffset:stdOffset);
localtime_r(&local, &timeinfo);
strftime(strftime_buf, sizeof(strftime_buf), "%m-%d-%Y", &timeinfo);  // generate MM-DD-YYYY local time character array
tft->setTextColor(TFT_BLUE, TFT_BLACK);
tft->drawString(strftime_buf, 50, 200, 4);
tft->setTextColor(TFT_GREEN, TFT_BLACK);
strftime(strftime_buf, sizeof(strftime_buf), "%I:%M:%S", &timeinfo); // generate HH:MM:SS local time character array
tft->drawString(strftime_buf, 5, 75, 7);
```
If there is a delay you can use two char arrays and generate both strings before displaying them.

I think something like that should work.

aviateur17:
> I think the RTC does expect it to be local time. I guess we ought to find out what the more "standard" way of operating is. It would make sense that a device such as a clock or watch would set to local time internally. Would a datalogger be similar or would it use UTC?

I started out using local time as the RTC time but I think it's more universal to use the UTC time and just convert to whatever standard you want to use.   I could see the datalogger data going either way.  It really depends upon how "global" you want the data to be.   After starting with local time I switched to UTC time because that is how it is utilized in NTP and I think the DST/STD conversion is way easier because you really do not have to fool with going backwards in time in the local time.   I have no idea what the Arduino standard is but in the building controls and industrial controls world I think UTC is the standard. Also, pilots use UTC as the standard but they are traversing between different time zones all the time.   I think an argument can be made either way but I think storing as UTC is better and calculating the offset to print the local time is the way to go.


> It seems like since you did all the work to make a nice DST routine, we should make the result a little more accessible to users. I have two ideas, let me know if either one seems good. One is to simply make the final offset (after DST adjustment) accessible through a function. The other is to just have two matching functions that return tm structs of either local or UTC.
> 
> What do you think? I may do option 2 either way. And we can also do neither or both.

I guess I would prefer a function that outputs local time in whatever manner is desired given my preference to have the underlying time in UTC.

timmbogner:
Yeah I see why UTC RTC makes everything easier. No problem with keeping it how it is. I realized my "function that returns the offset" idea is a bit superfluous. It would be good enough to just know if user is in DST or not... 

 **SO** what if you essentially just made checkDST() return a boolean so the user can have access to it? Then the user can make the logic to decide which offset macro to use to get the local time.

We could put a pin in the idea of functions that return tm structs for later. I'm not sure yet if that will be very complicated or useful.



_In unrelated news that I find sort of funny:_
Why didn't Espressif come up with a more unique acronym for [ESP-RTC](https://www.espressif.com/en/news/ESP-RTC)??😄

timmbogner:
OOPS, SORRY! I just noticed ```bool isDST;``` is global, and is p much what I described, right?

timmbogner:
I think ESP-NOW is pretty much solid. I can't get devices to receive a LoRa time, though. I've got Gateway 0x01 connected to MQTT and LoRa, then 0x03 is straight LoRa. 0x01 says it is sending, but I can't get the repeater or node to receive it. DataReadings and pings still make it okay.

Any ideas? I'm going to start working through it with DBGs to find where things are going wrong.

aviateur17:
> I think ESP-NOW is pretty much solid. I can't get devices to receive a LoRa time, though. I've got Gateway 0x01 connected to MQTT and LoRa, then 0x03 is straight LoRa. 0x01 says it is sending, but I can't get the repeater 0x03 or a node to receive it. DataReadings and pings still make it okay.
> 
> Any ideas? I'm going to start working through it with DBGs to find where things are going wrong.

Sorry, I've been neglecting this as I've been busy with other things.  Also been troubleshooting issues and have limited time.  I think LoRa is working on my code but I'm troubleshooting multiple instances of time being sent.   I'll submit what I have and then troubleshoot later.

aviateur17:
> I think ESP-NOW is pretty much solid. I can't get devices to receive a LoRa time, though. I've got Gateway 0x01 connected to MQTT and LoRa, then 0x03 is straight LoRa. 0x01 says it is sending, but I can't get the repeater 0x03 or a node to receive it. DataReadings and pings still make it okay.
> 
> Any ideas? I'm going to start working through it with DBGs to find where things are going wrong.

As far as LoRa goes the commit from a few days ago should work.  Are you using the latest commits in this branch? I'll send my output if I have time later on tonight but I'm not sure that I'll have time.

aviateur17:
Here is my debug output from:

ESPNOW -> LoRa gateway - address 0x22 (time received from 0x21):
```
Local date/time is: Fri Jun  9 07:07:53 2023 DST
Local date/time is: Fri Jun  9 07:08:53 2023 DST
Local date/time is: Fri Jun  9 07:09:53 2023 DST
Local date/time is: Fri Jun  9 07:10:53 2023 DST
Local date/time is: Fri Jun  9 07:11:53 2023 DST
Incoming ESP-NOW System Packet from 0x21
Received time via ESP-NOW from 0x21
ESP-NOW time master is 0x21
Received time via ESP-NOW from 0x21
Time adjust 1 secs
ESP-NOW time master is 0x21
Sending out time
Sending time via LoRa
Sending time to LoRa broadcast
Time adjust 0 secs
Transmitting LoRa message of size 11 bytes with CRC 0xa222 to LoRa MAC 0xffff
Sending time to LoRa Neighbor 1
Transmitting LoRa message of size 11 bytes with CRC 0x37ba to LoRa MAC 0xee23
Sending out time
Sending time via ESP-NOW
Sending time to ESP-NOW Peer 1
Sending time to ESP-NOW registered peers
Sending time via LoRa
LoRa airtime: 366782ms
Sending time to LoRa broadcast
Transmitting LoRa message of size 11 bytes with CRC 0xa222 to LoRa MAC 0xffff
Sending time to LoRa Neighbor 1
LoRa airtime: 366907ms
Transmitting LoRa message of size 11 bytes with CRC 0x37ba to LoRa MAC 0xee23
Sending time via ESP-NOW
LoRa airtime: 366987ms
Sending time to ESP-NOW Peer 1
Sending time to ESP-NOW registered peers
Incoming LoRa. Size: 11 Bytes, RSSI: -62.00dBm, SNR: 9.50dB, PacketCRC: 0x6db8, Total LoRa received: 0
We have received a ping request from 0xee23, Replying.
Transmitting LoRa message of size 11 bytes with CRC 0xfd98 to LoRa MAC 0xee23
LoRa airtime: 367248ms
```

LoRa -> LoRa Gateway - address 0x23 (time received from 0x22):
```
Incoming LoRa. Size: 11 Bytes, RSSI: -64.00dBm, SNR: 9.75dB, PacketCRC: 0x37ba, Total LoRa received: 0
Time adjust 1686312466 secs
Time change from STD -> DST
Sending out time
Sending time via LoRa
Transmitting LoRa message of size 11 bytes with CRC 0x6232 to LoRa MAC 0xffff
Time rcv from LoRa 0xee22
Transmitting LoRa message of size 11 bytes with CRC 0x6db8 to LoRa MAC 0xee22
LoRa ping sent to address: 0xee22
LoRa airtime: 255177ms
Incoming LoRa. Size: 11 Bytes, RSSI: -65.00dBm, SNR: 9.50dB, PacketCRC: 0xfd98, Total LoRa received: 1, CRC Ok Pct 100.00%
We have received a ping reply via LoRa from address ee22
LoRa Ping Returned: 295ms.
Time adj by 0 secs
Local date/time is: Fri Jun  9 07:12:45 2023 DST
Local date/time is: Fri Jun  9 07:13:45 2023 DST
```

timmbogner:
I've got it working on a different board. For some reason the TTGO LoRa32 v1 has an issue where it fully appears to transmit the time, but doesn't actually send anything. It's now running okay on the T3S3. The T3S3 seemed like it had issues at first, but dropped them and started working after some troubleshooting and I'm unclear what fixed them.

timmbogner:
Hey Jeff,
I am continuing to have issues with gateways not sending LoRa time. I can confirm that the same problem occurring on the LoRa32 is also occurring with an ESP8266 and SX1278 module.

Using T7-S3 and SX1278 it seems to send only one of the three LoRa time packets. This is similar to what the T3-S3 (with SX1262) did at first, though I'm not sure how I corrected it. I think this could be an issue with timing, and several receptions happening too quickly. The packet that it does receive is the one destined for neighbor#2, which is the last packet broadcasted.  

I think these are two separate issues, but I'm not positive. Could you share the the setup(s) that you have working? I have done some troubleshooting and cannot solve the first issue.

aviateur17:
> Hey Jeff, I am continuing to have issues with gateways not sending LoRa time. I can confirm that the same problem occurring on the LoRa32 is also occurring with an ESP8266 and SX1278 module.
> 
> Using T7-S3 and SX1278 it seems to send only one of the three LoRa time packets. This is similar to what the T3-S3 (with SX1262) did at first, though I'm not sure how I corrected it. I think this could be an issue with timing, and several receptions/transmissions happening too quickly. The packet that it does receive is the one destined for neighbor#2, which is the last packet broadcasted.

I've got a setup where I have a LoRa Neighbor 1 (N1) and no neighbor 2 and no LoRa sensor yet.   The N1 node doesn't reliably receive LoRa time periodically from the LoRa repeater that should be sending out the time. I'm not sure whether the issue is the transmitting repeater side or the receiving neighbor side yet.  I'll have to do more testing on this.  I'll add in some LoRa pings and add a LoRa sensor to determine what is happening.


> I think these are two separate issues, but I'm not positive. Could you share the the setup(s) that you have working? I have done some troubleshooting and cannot solve the first issue.

All of my LoRa nodes are using a Wemos ESP32 with the Hallard LoRa shield.


> EDIT: I think I've solved the second issue: In sendTimeLoRa(), each LoRa packet is sent to the radio without waiting for the previous to finish. I think the solution to this will be a similar setup to the asynchronous LoRa release function.

Thanks, I'll look for this and see if this is part of the problem.

timmbogner:
Greetings @aviateur17, that sounds awesome!

If I recall correctly, the ESP-NOW implementation was perfect, but there was a glitch in the LoRa portion. One thing I'd propose is that we start with a fresh pull request. This one is kind of a lot to go through, although mostly due to my own blunders.

An important upcoming change is going to be functionally major. I'm going to extend the 'data' portion of the DataReading structure to 5 bytes, and use CBOR to allow the data to be boolean, integer, binary, or float. This will only extend the DataReading to 8 bytes, which I think is pretty reasonable. There shouldn't be too many external changes for users. I will add new iterations of loadFDRS() for the different datatypes and then everything else will be handled internally. What do you think? SystemPackets will stay the same, so I'm not sure if it will affect your work here on timekeeping.

I have a rough sketch worked out with a lot of the experimentation done using this [TinyCBOR ](https://github.com/soburi/tinycbor/tree/master)library. I'll make it a priority to make some progress on it and post something today or tomorrow. 

Thanks!

aviateur17:
@timmbogner , agreed on posting a new pull request.  When I'm ready or close to ready and done with testing I'll post a new PR against the updated main branch.

Regarding the change in the structure of the DataReadings, in the LoRa code I'm using the length to differentiate between valid packets and invalid packets and then going in and dissecting what is actual data and what are the headers so once you have that change we'll need to go into that LoRa code and modify those parts.  Shouldn't be too difficult unless the size of the data is variable in nature.

PR #144, opened 2023-02-28
move #ifdefs used to enable or disable features from function scope t…

aviateur17:
…o file scope

Takes the #ifdefs used to disable and enable features like filesystem logging, ESP-NOW, LoRa, Debug, WiFi, etc. from function by function to including or excluding entire header files thus being a cleaner or more straightforward and simpler approach.

### DIFF
--- a/src/fdrs_checkConfig.h
+++ b/src/fdrs_checkConfig.h
@@ -79,8 +79,8 @@ void printLoggingInformation() {
 #else
 	DBG("log buffer delay in ms: NOT SPECIFIED - check config!");
 #endif
-#ifdef FS_FILENAME
-	DBG("log filename          : " + FS_FILENAME);
+#ifdef LOG_FILENAME
+	DBG("log filename          : " + LOG_FILENAME);
 #else
 	DBG("log filename          : NOT SPECIFIED - check config!");
 #endif
--- a/src/fdrs_gateway.h
+++ b/src/fdrs_gateway.h
@@ -45,17 +45,39 @@ uint8_t newCmd = cmd_clear;
 DataReading fdrsData[256]; // buffer for loadFDRS()
 uint8_t data_count = 0;
 
-#include "fdrs_oled.h"
+// Function Prototypes needed due to #ifdefs being moved outside of function definitions in header files 
+void broadcastLoRa();
+void sendLoRaNbr(uint8_t);
+void timeFDRSLoRa(uint8_t *);
+static uint16_t crc16_update(uint16_t, uint8_t);
+void sendESPNowNbr(uint8_t);
+void sendESPNowPeers();
+void sendMQTT();
+void sendLog();
+void resendLog();
+void releaseLogBuffer();
+
+#ifdef USE_OLED
+  #include "fdrs_oled.h"
+#endif
 #include "fdrs_debug.h"
-#include "fdrs_gateway_espnow.h"
-#include "fdrs_gateway_lora.h"
-#include "fdrs_gateway_wifi.h"
-#include "fdrs_gateway_filesystem.h"
-#include "fdrs_gateway_mqtt.h"
 #include "fdrs_gateway_serial.h"
 #include "fdrs_gateway_scheduler.h"
+#ifdef USE_ESPNOW
+  #include "fdrs_gateway_espnow.h"
+#endif
+#ifdef USE_LORA
+  #include "fdrs_gateway_lora.h"
+#endif
+#ifdef USE_WIFI
+  #include "fdrs_gateway_wifi.h"
+  #include "fdrs_gateway_mqtt.h"
+#endif
+#if defined(USE_FS_LOG) || defined(USE_SD_LOG)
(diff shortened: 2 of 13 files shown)


src/fdrs_gateway_filesystem.h
```diff
@@ -1,34 +1,23 @@
...
 #endif
 #ifdef USE_FS_LOG
 #include <LittleFS.h>
 #endif
-#if defined(USE_SD_LOG) || defined(USE_FS_LOG)
 #include <time.h>
-#endif
-#if defined(USE_SD_LOG) || defined(USE_FS_LOG)
+
+#define SD_MAX_FILESIZE 1024
```
aviateur17, line 11: SD_MAX_FILESIZE was not previously defined.  Not sure that previously the code would compile without this defined.

src/fdrs_gateway_filesystem.h
```diff
@@ -1,34 +1,23 @@
...
+#define FS_MAX_FILESIZE 1024
```
aviateur17, line 12: FS_MAX_FILESIZE was not previously defined.  Not sure that previously the code would compile without this defined.

src/fdrs_gateway_espnow.h
```diff
@@ -228,86 +223,81 @@ void pingback_espnow()
...
-    }
-    esp_now_send(ESPNOW2, (uint8_t *)&thePacket, j * sizeof(DataReading));
-    esp_now_del_peer(ESPNOW2);
+      esp_now_send(ESPNOW2, (uint8_t *)&thePacket, j * sizeof(DataReading));
+      esp_now_del_peer(ESPNOW2);
 
-    break;
-  }
+      break;
+    }
```
aviateur17, line 295: Just added a tab to the previous code for proper identation.

src/fdrs_gateway.h
```diff
@@ -193,3 +215,16 @@ void loopFDRS()
   }
 }
 
+// "Skeleton Functions related to FDRS Actions"
```
aviateur17, line 218: Skeleton Functions needed to allow these functions to be defined in ACTIONS even when the features are not enabled.   Example:  If LoRa is not enabled that the entire file that has the LoRa functions is excluded from compilation but actions still include broadcastLoRa so need that function present for compilation to succeed.

src/fdrs_checkConfig.h
```diff
@@ -79,8 +79,8 @@ void printLoggingInformation() {
 #else
 	DBG("log buffer delay in ms: NOT SPECIFIED - check config!");
 #endif
-#ifdef FS_FILENAME
-	DBG("log filename          : " + FS_FILENAME);
+#ifdef LOG_FILENAME
```
aviateur17, line 82: I don't believe FS_FILENAME existed in the previous code

src/fdrs_globals.h
```diff
@@ -17,7 +17,8 @@
 // MQTT Topics
 #define TOPIC_DATA    "fdrs/data"
 #define TOPIC_STATUS  "fdrs/status"
-#define TOPIC_COMMAND "fdrs/command" 
+#define TOPIC_COMMAND "fdrs/command"
+#define TOPIC_DATA_BACKLOG "fdrs/databacklog"   // Used in filesystem module
```
aviateur17, line 21: I don't believe TOPIC_DATA_BACKLOG was defined in prior code.

aviateur17:
After modifying the code I went though and made sure all examples compiled using Arduino 1.8.19.  I went through the config files and commented and uncommented the USE_xxx features out and compiled different combinations to make sure all the code works.

timmbogner:
Awesome, thanks! That's gonna make things a lot simpler. I'll get this in before the release, which I'm hoping to do tomorrow.

timmbogner:
ESP8266 does *not* like your ```const``` changes in ESP-NOW.

aviateur17:
Thanks for testing.  I tested against ESP32 board only.   Guess I will have to test ESP8266 in the future.  I've been wondering if there are some automated compile tests that can be done to make sure changes compile against all configurations and all platforms.  I'm not sure if that's called unit testing or what that automated testing would be called.  I gotta believe there is a way to set that up.

aviateur17:
After I finish making these timekeeping changes I'll go back and compile against ESP8266 and update this PR.

timmbogner:
No rush... I'm not going to finish up tonight so I won't get around to pulling it until tomorrow afternoon.

>I gotta believe there is a way to set that up.

Maybe with Arduino CLI and a batch file? That would be really handy. 

Actually... I should look into that in general to streamline my process of editing in the git folder, drag/copy to Arduino folder, click confirm, then clicking compile in Arduino IDE. It's not too bad til you have to fix three small errors in a row and suddenly want to give up 😅

aviateur17:
Fixed issue with const declaration in commit https://github.com/timmbogner/Farm-Data-Relay-System/pull/144/commits/9b5323e447870e7672bccaa8290489dcfc72594d

PR #151, opened 2023-03-08
FDRS OLED features

aviateur17:
Add OLED 'pages' and blank OLED screen

Adding "paging" feature to OLED such that different sets of data can be shown.  First page will be standard debug data.  Rest of pages will be defined later.  #define in config file states how long each page will stay on screen (default 30 seconds).  When any debug information is added screen immediately switches to that debug screen until there is no action for a specified amount of time (default 30 seconds).

### DIFF
--- a/src/fdrs_oled.h
+++ b/src/fdrs_oled.h
@@ -1,9 +1,12 @@
 
 
 #include <ESP8266_and_ESP32_OLED_driver_for_SSD1306_displays/src/SSD1306Wire.h>
+#define DISPLAY_PAGES 4
 
 String debug_buffer[5] = {"", "", "", "", ""};
 SSD1306Wire display(0x3c, OLED_SDA, OLED_SCL); // ADDRESS, SDA, SCL
+unsigned long displayEvent = 0;
+uint8_t displayPage = 0;
 
 void draw_OLED_header()
 {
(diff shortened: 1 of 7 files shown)


aviateur17:
Future enhancements could include but not limited to adding additional "pages" and use library that does not have to be modified.

src/fdrs_oled.h
```diff
@@ -33,30 +36,102 @@ void draw_OLED_header()
     display.setTextAlignment(TEXT_ALIGN_LEFT);
     display.setFont(ArialMT_Plain_10);
     #endif
+    display.drawHorizontalLine(0, 15, 128);
+    display.drawHorizontalLine(0, 16, 128);
```
timmbogner, line 40: Note: On dual-color OLEDs, line 15 is yellow and 16 is blue. Logically, line 16 should be drawn in the page since it's the beginning of the large section of the display. /nitpick

timmbogner:
Quick question: in VSCode, I like to use the autoformatter to make my messy code look better. It currently changes any occurrence like this:
```
void foo(){
  bar();
}
```
to 
```
void foo()
{
  bar();
}
```
I prefer the former, but what can I change in VScode to make it do it right?

aviateur17:
Agreed, I prefer the more condensed code as far as brackets go.  I've never used the autoformatter but I'll spend some time researching and let you know what I find.

aviateur17:
Looks like you might need an add-on called prettier:
https://www.reddit.com/r/vscode/comments/115h9d1/how_to_change_curly_brackets_behavior_on_file_save/

https://prettier.io/docs/en/rationale.html#empty-lines

aviateur17:
This PR should be complete.  I don't plan to add any other pages for awhile.   I wanted to get the "blank page" stuff in there primarily.

timmbogner:
Merging this and OTA will be my first task once I get back to work on things. I'm a little behind on things and splitting time with my day-job now.

aviateur17:
No worries, take your time. I know you're busy.

timmbogner:
Not sure if I said this elsewhere, but this is nifty and I appreciate it!

PR #172, opened 2023-05-15
Update fdrs_node_lora.h

sbobowski:
Wrap 'yield();' in '#ifdef ESP8266' as it does not appear to be needed by any other MCUs


timmbogner:
Looks good, thanks a lot! Definitely drop in if you run into any more CubeCell issues.

PR #193, opened 2024-02-10
FDRS Support for timekeeping - 2024 version

aviateur17:
Add gateway time keeping, NTP, RTC support

For Gateways, add time keeping in all nodes - MQTT, serial, ESP-NOW, LoRa.  
Add support to fetch time via NTP if WiFi is enabled.

Add support for I2C RTC devices - DS3231 and DS1307 and compatible

Gateway will only send time out via serial if WiFi or RTC is in use to prevent loop.  Don't want serial node to send time back to WiFi/MQTT node.

Priority in Time master Serial > ESP-NOW > LoRA which means a gateway that is receiving time via serial will reject time via ESP-NOW and LoRA.  A gateway that is receiving time via ESP-NOW will reject time via LoRa.   If no time is received after 60 minutes then the "master" time server will reset to none. 

Automatic transition between daylight savings time and standard time for both US and Europe.

Still need to work on support for controllers and sensors in a future merge
Will also verify that compilation works in Arduino environment (I am using VSCode and PlatformIO)
Also need to verify compilation against different platforms such as ESP8266, raspberry pi Pico, SAMD21.  What other micro-controllers are intended to be supported?


src/fdrs_gateway.h
```diff
@@ -118,6 +120,12 @@ void beginFDRS()
   Serial.begin(115200);
   UART_IF.begin(115200, SERIAL_8N1, RXD2, TXD2);
 #endif
+#if defined(USE_OLED) || defined(USE_DS3231) || defined(USE_RTC_DS1307)
```
aviateur17, line 123: Typo - need to change USE_DS3231 to USE_RTC_DS3231 here.

src/fdrs_gateway_lora.h
```diff
@@ -410,33 +414,20 @@ crcResult getLoRa()
               transmitLoRa(&sourceMAC, &pingReply, 1);
             }
           }
-          else
```
aviateur17, line 413: Need to review why these lines were removed - verify this is intended.

src/fdrs_gateway_lora.h
```diff
@@ -410,33 +414,20 @@ crcResult getLoRa()
...
+              adjTimeforNetDelay(netTimeOffset);
+              timeMaster.tmLastTimeSet = millis();
             }
-            else if (receiveData[0].param == 0)
-            {
-              DBG("We have received a ping request from 0x" + String(sourceMAC, HEX) + ", Replying.");
-              SystemPacket pingReply = {.cmd = cmd_ping, .param = 1};
-              transmitLoRa(&sourceMAC, &pingReply, 1);
+            else {
+              DBGF("LoRa 0x" + String(sourceMAC, HEX) + " is not time master, discarding request");
```
aviateur17, line 430: Should be DBG and not DBGF

src/fdrs_gateway_lora.h
```diff
@@ -626,4 +617,72 @@ crcResult handleLoRa()
...
+
+// FDRS Sensor pings gateway and listens for a defined amount of time for a reply
+// Blocking function for timeout amount of time (up to timeout time waiting for reply)(IE no callback)
+// Returns the amount of time in ms that the ping takes or predefined value if ping fails within timeout
+uint32_t pingFDRSLoRa(uint16_t *address, uint32_t timeout)
+{
+    SystemPacket sys_packet = {.cmd = cmd_ping, .param = 0};
+
+    transmitLoRa(address, &sys_packet, 1);
+    DBGF("LoRa ping sent to address: 0x" + String(*address, HEX));
```
aviateur17, line 651: DBG and not DBGF

timmbogner:
Well I'll confess, the CBOR thing has more facets than I first considered. I think that will be on hold for at least a week while I think, fyi.
I'll give this a try in a bit!

aviateur17:
I'm not sure what CBOR is.  Haven't look into it yet and haven't heard the guy with a Swiss accent reference it yet.

timmbogner:
From the end of the previous timekeeping PR:
>An important upcoming change is going to be functionally major. I'm going to extend the 'data' portion of the DataReading structure to 5 bytes, and use CBOR to allow the data to be boolean, integer, binary, or float. This will only extend the DataReading to 8 bytes, which I think is pretty reasonable. There shouldn't be too many external changes for users. I will add new iterations of loadFDRS() for the different datatypes and then everything else will be handled internally. 

It's nothing too exciting, just a data format. I'd been planning to try to hammer it out this weekend and just letting you know that this isn't happening so soon.

aviateur17:
For sensor nodes that power up, collect sensor data, send sensor data, power down, no need for them to know the date and time, right?

timmbogner:
>For sensor nodes that power up, collect sensor data, send sensor data, power down, no need for them to know the date and time, right?

Right. I don't see why it would be required there. If the user runs `addFDRS()` and makes it a controller, then would be were timekeeping comes into play I think.

I'm still reading through the new code, but I think it's great so far. I forget how it was before exactly, but the new datatypes and time master concept look like a good move.

aviateur17:
@timmbogner  question for you in the scenario where we have a controller using ESP-NOW.  It seems that the controller registers itself to the gateway when we call the addFDRS() function.  This function is used to register a call back function in case we want to send data to the controller.   However, what if we have a controller where we do not need to register a call back function?  Maybe this is really the definition of a controller where addFDRS is really required but I'm thinking that maybe we decouple the registration part from the addFDRS function call.   Every controller should register to the gateway even if addFDRS is not called, if I understand correctly so maybe better just to register automatically even if addFDRS is not called in main.cpp.  So if you are in agreement then all addFDRS would do would register the callback function.  The registration of the controller with the gateway would be called independent of addFDRS.  

Additionally, if the registration with the gateway fails, I would think that we would want to periodically try to reregister with the gateway.

Then I would wonder if we have a sensor node that does not go to sleep, we should probably register the sensor with the gateway and thus the gateway would send the sensor the time which makes sense to me if the sensor is continuously powered on.  What are your thoughts on that?

Finally, in the "node" code there is quite a bit of ESP-NOW and LoRa stuff in the fdrs_node.h file that I think can be moved to the fdrs_node_espnow or fdrs_node_lora files.  If you agree, I think I'll take up organizing that code but in a different PR.

I think the ESP-NOW controller  code is good so now to work on LoRa controller...

timmbogner:
>However, what if we have a controller where we do not need to register a call back function? Maybe this is really the definition of a controller where addFDRS is really required

You kind of called it right there. If a node is receiving data, it will need a callback in which to handle it. If it needs a callback then it will need to be defined somewhere. Thus, `addFDRS()` is essential to being a controller. In ESP-NOW it does other things too, of course. 

I see now that in LoRa, `addFDRS()` really does very little. While it would be possible to  perhaps introduce another way of setting the callback just for LoRa, I think it would be better to keep ESP-NOW and LoRa configuration and usage as similar as possible for now. I will keep thinking about it, though.
#
In nodes, `gtwy_timeout` is a variable that stores the frequency at which to refresh and retain their spot on the gateway's peer list. It is initialized to 300000 (5 mins). When nodes connect to a gateway, they receive a value to store as `gtwy_timeout`. `addFDRS()` sets `is_controller` to 'true', and that enables a registration to happen at the interval in `gtwy_timeout`. So what I'm getting at is that the controller *should* try to connect again after 5 minutes, as long as addFDRS() has been called.
#
>Then I would wonder if we have a sensor node that does not go to sleep, we should probably register the sensor with the gateway and thus the gateway would send the sensor the time which makes sense to me if the sensor is continuously powered on. What are your thoughts on that?

I think in this case, if the user wanted the sensor to get the time, they could register as a controller. I do see how it might be nice to avoid requiring a registration spot just for a clock, though. I still need to look deeper into your code, so take all of this with a grain of salt. I haven't fully wrapped my head around it at this time, but that's just because I've been distracted, not anything to do with the code itself. I'll give it attention today for sure.
# 
>Finally, in the "node" code there is quite a bit of ESP-NOW and LoRa stuff in the fdrs_node.h file

Agreed. I'm not sure if that stayed there for a reason or not, but `addFDRS()` can definitely move to node_espnow if there are no conflicts.

timmbogner:
Jumping in to testing. I'm using Arduino IDE 2.3.0 and ESP32 board defs 2.0.14:

0_MQTT_Gateway works great, gets the time.

1_UART_Gateway with ESP-NOW gives the following error:
```
...\Farm-Data-Relay-System\src/fdrs_gateway_espnow.h: In function 'esp_err_t sendTimeESPNow()':
...\Farm-Data-Relay-System\src/fdrs_gateway_espnow.h:397:26: error: invalid conversion from 'uint8_t*' {aka 'unsigned char*'} to 'uint8_t' {aka 'unsigned char'} [-fpermissive]
  397 |     result1 = sendESPNow(ESPNOW1, &sys_packet);
```
1_UART_Gateway with only LoRa enabled gives the following error:
```
...\Farm-Data-Relay-System\src/fdrs_gateway_lora.h: In function 'uint32_t pingFDRSLoRa(uint16_t*, uint32_t)':
...\Farm-Data-Relay-System\src/fdrs_gateway_lora.h:653:5: error: 'pingFlagLoRa' was not declared in this scope; did you mean 'pingFDRSLoRa'?
  653 |     pingFlagLoRa = false;
  ```

**Comment by aviateur17** [comment]
If the callback function was fixed at a certain name then can't we do all of that automatically without having them call the addFDRS() function?  Then all we would have to do is check if the node was a controller or a sensor.   Could we just call a node that doesn't define DEEP_SLEEP as a controller and one that does define DEEP_SLEEP as a sensor?  Then use that as a condition for registering the callback function and registering.   It seems at this time there can only be one callback function.  Maybe you've thought of this?

Regarding the fdrs_node.h code I may take that reorganization as a task at a later point.

**Comment by timmbogner** [comment]
>It seems at this time there can only be one callback function. Maybe you've thought of this?

I've been thinking about it more since you mentioned it and you're right, it could be static. I think a problem with what you say is that each ESP-NOW device that registers as a controller takes up one slot in the peer list, which is limited to 16. It would be best to keep these to the devices that actually need to receive data. Shouldn't there be a way for a node to request the time? I guess this is sort of what you asked about early on. I think the way it should work is: If a node connects as a controller, it will get time updates from the gateway. Any sensor can send a request for the time to its gateway and get the time that way, but it won't be fed the time automatically. 

In general, I think a sensor should not be listening or receiving anything unless the user explicitly runs `addFDRS()` (or `pingFDRS()` etc). Just a thought I came up with now: For a user to request the time from a gateway what about `queryFDRS(enum)` and we can start an enum with time and other things the node may want to query a gateway about (I'll think of other things later :D) It could end up as a handy diagnostic tool.

As a side note: The controller callback is going to come under more scrutiny soon, as I trudge through this CBOR stuff. If the incoming data can be one of several types, will there have to a callback for each? Or will there a function within a single callback to decode it? Will there be more variables passed to the callback? It doesn't help that despite using them a few times, I still need to review every time I work with function pointers.

#

**I've tested ESP-NOW Gateways 00 -> 01 ->02 plus a controller, and they all are working perfectly and keeping time. LoRa next.**

**Comment by aviateur17** [comment]
> It would be best to keep these to the devices that actually need to receive data. 

Sounds good. makes perfect sense.
 
>. Any sensor can send a request for the time to its gateway and get the time that way, but it won't be fed the time automatically.

I'll look into this.  Shouldn't be hard to do this.  I think the code is pretty much there.  Just need a good function call.

> As a side note: The controller callback is going to come under more scrutiny soon, as I trudge through this CBOR stuff. If the incoming data can be one of several types, will there have to a callback for each? Or will there a function within a single callback to decode it? Will there be more variables passed to the callback? It doesn't help that despite using them a few times, I still need to review every time I work with function pointers.

Sounds good.  I'll leave that stuff alone.

> **I've tested ESP-NOW Gateways 00 -> 01 ->02 plus a controller, and they all are working perfectly and keeping time. LoRa next.**

I have to do some more testing in edge cases and different settings so I'll post more updates as I find issues.

Sounds like we're going to make the assumption that the MQTT gateway is the only gateway that would be the initial source of time, right?

**Comment by aviateur17** [comment]
So, using your example, let's say 00 does not have Internet access but does have RTC.  Somehow, the time on the RTC is off and not correct.  02 has GPS.  GPS should be preferred over RTC.   Do we implement a check if the two time sources differ use the one that should be more accurate?  01 would be able to tell if the time they are receiving from 00 and 02 are different by some appreciable amount.

**Comment by aviateur17** [comment]
> So, using your example, let's say 00 does not have Internet access but does have RTC. Somehow, the time on the RTC is off and not correct. 02 has GPS. GPS should be preferred over RTC. Do we implement a check if the two time sources differ use the one that should be more accurate? 01 would be able to tell if the time they are receiving from 00 and 02 are different by some appreciable amount.

I'm thinking that we make the assumption that if a gateway has an RTC that it should be updated at fairly regular intervals by some Internet connection or GPS otherwise why put the RTC on that gateway.   So I'm going to then assume that the RTC and other gateway with GPS are assumed close enough in time to each other.  If there is an issue where the time is off (> 10 seconds?) then raise a DBG message and hope that someone sees the message.  Thoughts on that?

**Comment by timmbogner** [comment]
Yeah I'd been thinking about RTC too.  I think I may have come up with something just now. What if the RTC gateway never shares its time until a time is sent to it? Basically we assume the RTC time is incorrect until an outside time packet sets it. If the user wants to pre-set their RTC and use it as the only time source, we can set up a config to override this behavior.

So... what am I forgetting?

**Comment by timmbogner** [comment]
> I'm thinking that we make the assumption that if a gateway has an RTC that it should be updated at fairly regular intervals by some Internet connection or GPS otherwise why put the RTC on that gateway. So I'm going to then assume that the RTC and other gateway with GPS are assumed close enough in time to each other. If there is an issue where the time is off (> 10 seconds?) then raise a DBG message and hope that someone sees the message. Thoughts on that?

I didn't see this til just now. I think we're thinking pretty closely. I had also thought about a routine that keeps track of whether incoming times are synced. I like the idea of a debug message, and we could expand from there if wanted.

**Comment by aviateur17** [comment]
Periodically generate a FDRS message indicating a time issue so it can be sent to the front end.  Like an alarm condition.

**Comment by timmbogner** [comment]
> Periodically generate a FDRS message indicating a time issue so it can be sent to the front end. Like an alarm condition.

Excellent idea! We can use `INTERNAL_ACT` to send it.

File: src/fdrs_gateway_serial.h [hunk 5]
```diff
@@ -48,19 +52,48 @@ void getSerial() {
     }
     ln = s;
     newData = event_serial;
-    DBG("Incoming Serial.");
-
+    DBG("Incoming Serial: DR");
+      String data;
+      serializeJson(doc, data);
+      DBG("Serial data: " + data);
```
timmbogner, line 58: JSON data output to the console should be displayed regardless of FDRS_DEBUG. Among other things, it's documented or mentioned somewhere to work that way, so we shouldn't change it.

timmbogner:
Ayyo, now it's working! Had to switch to using `radio.transmit()` instead of `radio.startTransmit()`. 

My setup is 00_MQTT->01_Serial->ESP-NOW->02_ESP-NOW/LoRa->03_LoRa

I'll be playing with it alot more in the next few days, plus I've gotta bust out the ol' T-Watch. Very excited!

src/fdrs_gateway_serial.h
aviateur17: Timm, I'm not sure I understand that comment.  Can you elaborate a bit on:
1." JSON data output to the console should be displayed regardless of FDRS_DEBUG."
2. The idea is that the console interface always doubles as a data interface.

Are you saying that even if the DBG is not there we should see the data on the console?

aviateur17:
> Ayyo, now it's working! Had to switch to using `radio.transmit()` instead of `radio.startTransmit()`.

My code is working with:
```
int state = radio.startTransmit(pkt, sizeof(pkt));
```

Wonder what's going on.

timmbogner:
>My code is working with:

That's interesting, does it still work for you with my change?

>Wonder what's going on.

It must be a timing thing. This low-level LoRa is tricky. What hardware are you on?

When sending a system packet with transmitLoRa()`, `startTransmit()` is called it and begins the transmission, we set `transmitFlag`, and it moves on in code.  Within `handleLoRa()`, we wait until the interrupt is triggered, then if `transmitFlag` is set it runs `finishTransmit()`. Next, it checks if there is an async LoRa release happening. This is a potential conflict, but we'll overlook it for now. There usually shouldn't be, so it goes to this:
```cpp
DBG("LoRa airtime: " + String(millis() - tx_start_time) + "ms");
radio.startReceive(); // return to listen mode
enableInterrupt = true;
transmitFlag = false;
``` 
The airtime display assumes that it is at the end of an async release, so it never has accurate data. 
#
My guess as to the issue is that the interrupt isn't triggering where we expect it to, and it never has a chance to `finishTransmit()`. When switched to transmit(), all of the interrupt stuff happens before it moves on, so it doesn't have a chance to get confused.

I'm going to fill the handleLoRa() function with DBGs and try to get to the bottom of it. I'm hoping `transmit()` will be okay for now. I've had a post-it on my monitor that says "investigate blocking transmit" since the last timekeeping effort. I think this is why I couldn't get it to work last time too, but I never quite made the connection.

src/fdrs_gateway_serial.h
timmbogner: Yes. Even if `FDRS_DEBUG` is disabled, the "Serial" (as well as "Serial1") interface should print out DataReadings when `sendSerial()` is called. I also think it best if they are delivered surrounded by line-breaks. Just in case a front end application is ingesting from that interface and needs it more standardized. Does that help?

It was originally implemented due to the ESP8266 having only one serial interface. If you need/want to re-use that interface for GPS or something, maybe we could have a way to explicitly disable this. Afterthought: A way to fully disable the serial interfaces might be called-for if we need that.

Side note: I might disable the new indentations in the debug console when at level 0.  The blank space is just a bit jarring (probably just to me) but it's also easier to read when info is closer to the timestamp.

timmbogner:
I added some debug messages and the transmissions began working. I believe I found the problem.

In `sendTimeLoRa()` you run `transmitLoRa()` twice in a row with no delay. If you are using `radio.startTransmit()`, you need to wait for the first operation to finish before doing it again. Probably just as easy to use `radio.transmit()`.

You can probably recreate the issue by disabling FDRS_DEBUG.

timmbogner:
I assembled some console recordings to demonstrate. `*INT*` is printed when the interrupt itself is called. The rest of the new debug messages are either commented out in the existing `handleLoRa()` code or self-explanatory.

First, this is sending the time with radio.startTransmit(). Note how the interrupt is triggered right after it (tried to) send the second message. That was the chip indicating it was finished with the first transmit:
``` cpp
12:12:16.838 -> Sending out time
12:12:16.838 -> Sending time via LoRa
12:12:16.838 -> Sending time to LoRa broadcast
12:12:16.838 -> Transmitting LoRa SYSTEM message of size 11 bytes with CRC 0x9649 to LoRa MAC 0xffff
12:12:16.877 -> Sending time to LoRa Neighbor 2
12:12:16.877 -> Transmitting LoRa SYSTEM message of size 11 bytes with CRC 0xc248 to LoRa MAC 0xee03
12:12:16.877 -> *INT*
12:12:16.877 -> Sending time to ESP-NOW Peer 1
12:12:16.877 -> Sending time to ESP-NOW Peer 2
12:12:16.877 -> Sending time to ESP-NOW registered peers
12:12:16.877 -> Interrupt triggered
12:12:16.877 -> TxFlag: 1
12:12:16.877 -> TxStatus: 3
12:12:16.877 -> Finishing Transmit
12:12:16.877 -> *INT*
12:12:16.877 -> LoRa airtime: 60038ms
```

Next is the way it works with `radio.transmit()`. I was curious how this would play out. It doesn't have a chance to get to `handleLoRa()` while sending the two packets, so the interrupt flag is only handled once, after everything else is finished. It luckily handles it by running radio.finishTransmit(), which is not needed here, but I don't think it hurts anything:

``` cpp
12:07:28.567 -> Sending out time
12:07:28.567 -> Sending time via LoRa
12:07:28.567 -> Sending time to LoRa broadcast
12:07:28.567 -> Transmitting LoRa SYSTEM message of size 11 bytes with CRC 0xadc9 to LoRa MAC 0xffff
12:07:28.633 -> *INT*
12:07:28.633 -> Sending time to LoRa Neighbor 2
12:07:28.633 -> Transmitting LoRa SYSTEM message of size 11 bytes with CRC 0xf9c8 to LoRa MAC 0xee03
12:07:28.666 -> *INT*
12:07:28.666 -> Sending time to ESP-NOW Peer 1
12:07:28.666 -> Sending time to ESP-NOW Peer 2
12:07:28.666 -> Sending time to ESP-NOW registered peers
12:07:28.666 -> Interrupt triggered
12:07:28.666 -> TxFlag: 1
12:07:28.666 -> TxStatus: 3
12:07:28.666 -> Finishing Transmit
12:07:28.666 -> LoRa airtime: 240116ms
```

Here is a look at a DataReading being sent under this paradigm:
``` cpp
12:13:58.810 -> Incoming ESP-NOW.
12:13:58.810 -> Sending to ESP-NOW Neighbor #2
12:13:58.810 -> Sending to ESP-NOW peers.
12:13:58.851 -> Sending to LoRa neighbor buffer
12:13:58.851 -> Sending to LoRa broadcast buffer
12:14:01.917 -> Transmitting LoRa message of size 13 bytes with CRC 0x858f to LoRa MAC 0xee03
12:14:01.987 -> *INT*
12:14:01.987 -> Interrupt triggered
12:14:01.987 -> TxFlag: 1
12:14:01.987 -> TxStatus: 2
12:14:01.987 -> Finishing Transmit
12:14:01.987 -> Transmitting LoRa message of size 13 bytes with CRC 0xea8f to LoRa MAC 0xffff
12:14:02.032 -> *INT*
12:14:02.032 -> Interrupt triggered
12:14:02.032 -> TxFlag: 1
12:14:02.032 -> TxStatus: 3
12:14:02.032 -> Finishing Transmit
12:14:02.032 -> LoRa airtime: 92ms

**Comment by aviateur17** [comment]
Sounds good.  I probably won't have a chance to look at the information or analyze it for a few days or a week.  I'm trying to finish up the timekeeping for the node ESP-NOW and LoRa.  I also need to look at your comments on JSON console output as I still don't understand those either.  
In my testing with this code I've ran into a couple of issues that I keep in the back of my mind:
1. DBG or serial output takes quite a long time to do, relatively speaking, and the placement of the console output isn't necessarily in line with what is happening in the code.  The console output could appear after several other lines of code have executed.
2. Order of the network calls (ESP-NOW and LoRa Xmit and Receive) matters relative to the code around it.  There was some code for ESP-NOW that transmitted and then did some stuff like DBG and then waited for a reply.  Well the reply came before the DBG happened so by the time the code went to wait for a reply the reply had already been received and so the code waiting for a reply timed out.  Moving the ESP-NOW Transmit right before the spinning loop waiting for the reply resolved the issue.  So the order of the code matters quite a bit and console output takes a considerable amount of time.

**Comment by aviateur17** [comment]
[172f61c](https://github.com/timmbogner/Farm-Data-Relay-System/pull/193/commits/172f61ccb1b20426eb908105d729f1883926c153)



All: Sensors can request gateway to send time
All: Time source tracking per gateway/controller/sensor and prioritizing sources
All: Time difference macro to simplify comparing the last time some event happened:
#define TDIFF(prevMs,durationMs) (millis() - prevMs > durationMs)

Gateway: Lora asynchronous ping support so CPU time is not wasted while waiting for a ping response.

Gateway: Add support for GPS module (AT6558 based module) If others have GPS modules that receive different 

Gateway: Clean up loopFDRS() so that the whole loop is consistent: move action handling into separate function and rename updateTime to handleTime.

Controller/Sensor: ESP-NOW ping is now asynchronous.  Probably not required since pings are on the order of 50ms or less.

Smaller enhancement of DBG statements

File: src/fdrs_gateway_espnow.h [hunk 6]
```diff
@@ -470,15 +472,19 @@ void sendESPNow(uint8_t address)
...
-  if(timeMaster.tmType == TM_NONE || timeMaster.tmType == TM_LORA || (timeMaster.tmType == TM_ESPNOW && timeMaster.tmAddress == incMAC[4] << 8 | incMAC[5])) {
+  if(timeMaster.tmNetIf <= TMIF_ESPNOW ) {
     DBG("Received time via ESP-NOW from 0x" + String(incMAC[5], HEX));
-    if(timeMaster.tmAddress == 0x0000) {
-      timeMaster.tmType = TM_ESPNOW;
-      timeMaster.tmAddress = incMAC[4] << 8 & incMAC[5];
-      DBG("ESP-NOW time master is 0x" + String(incMAC[5], HEX));
+    if(timeMaster.tmNetIf < TMIF_ESPNOW) {
+      timeMaster.tmNetIf = TMIF_ESPNOW;
+      timeMaster.tmAddress = incMAC[4] << 8 | incMAC[5];
```
timmbogner, line 479: I'm not great with bitwise operations yet, could you save me some research and tell me what `incMAC[4] << 8 | incMAC[5]` is expressing in English?

In the following, I understand the second condition is that the address is not 0x00, but what is the fist check for?
```if((timeMaster.tmAddress != (ESPNOW1[4] << 8 | ESPNOW1[5])) && ESPNOW1[5] != 0x00) {```

Thanks, sorry for the basic question.

File: src/fdrs_gateway_wifi.h [hunk 7]
```diff
@@ -175,11 +175,18 @@ void begin_wifi()
...
   {
-    DBG("Connecting to WiFi...");
-    DBG(FDRS_WIFI_SSID);
-    delay(500);
+    connectTries++;
+    DBG("Connecting to WiFi SSID: " + String(FDRS_WIFI_SSID) + " try number " + String(connectTries));
+    delay(1000);
+    WiFi.reconnect();
+    if(connectTries >= 15) {
+        DBG("Restarting ESP32: WiFi issues\n");
```
timmbogner, line 186: Awesome and useful

aviateur17:
Hey Timm, I feel like I'm getting close to completing this functionality.  I'm finishing up my added features testing and then I need to confirm successful compilation and go through all of the DBGs.   
On the DBGs, what is your thought on which DBG statements you want at level 0? Should I leave that up to you, should I take a stab at guessing, or can you give some examples of what you want at level 0?
Regarding the compilation, there are so many combinations of #defines now that it's tough to be able to test every possible one.  I'll try to test the most common ones for sure.

timmbogner:
This is all fantastic! Sorry I've been a little unresponsive. I haven't been able to test the newest commit because of some issues with my WiFi. I think the router I bought used on Craigslist in 2016 might have finally failed. I presently realized that's a good excuse to test ethernet, which I'll work on today and tomorrow.

For DBGs: I think I'd say you should set up LoRa and any other messages how you like them, and I'll look at that as a rough guide for the rest. Let me know any feedback once I do. I had set some of them up for ESP-NOW in the pull request, (now in main) that aren't here in this one. Assuming they'll show up when you resolve the conflicts. 

I needed to define `GPS_TXD` and `GPS_RXD` for it to compile in Arduino. I'll begin testing various things today after I get ethernet running.  I noticed after I commented on the new reboot code for ESP32 that it would probably need protection from being compiled in 8266 too. Generally, though, this is mind-blowingly cool.

aviateur17:
> This is all fantastic! Sorry I've been a little unresponsive. I haven't been able to test the newest commit because of some issues with my WiFi. I think the router I bought used on Craigslist in 2016 might have finally failed. I presently realized that's a good excuse to test ethernet, which I'll work on today and tomorrow.

No worries on any delays.  Totally understand that everyone is busy with many things.

> 
> For DBGs: I think I'd say you should set up LoRa and any other messages how you like them, and I'll look at that as a rough guide for the rest. Let me know any feedback once I do. I had set some of them up for ESP-NOW in the pull request, (now in main) that aren't here in this one. Assuming they'll show up when you resolve the conflicts.

I will look at the DBGs for ESP-NOW and try to replicate for LoRa.  My thought, in general, is that DBGs are also meant to be compatible with OLED as far as length (you said this quite awhile ago) and should be very basic in nature and not too many of them included.

> 
> I needed to define `GPS_TXD` and `GPS_RXD` for it to compile in Arduino. I'll begin testing various things today after I get ethernet running. I noticed after I commented on the new reboot code for ESP32 that it would probably need protection from being compiled in 8266 too. Generally, though, this is mind-blowingly cool.

I'll check on the #defines for those.  I thought I had them surrounded by USE_GPS but I could be wrong.  I'll go back and check.

timmbogner:
Ethernet seems to work out-of-the-box! 

I haven't fully digested all of the code and the various rules and expected behaviors yet. As such, I'm still not sure what the special rule about sending time to gateways addressed 0x00 does. Does the case of my MQTT/NTP gateway being 0x01 throw that off? I think that rule is the only thing I'm uneasy about so far. The order of gateways can't be assumed by the system, and it should be able to work the same way as if the addresses were reversed (like if the highest address was closest to the front-end) for example. Once I get my head around things I'll try to make some constructive suggestions for alternatives.

aviateur17:
> As such, I'm still not sure what the special rule about sending time to gateways addressed 0x00 does. Does the case of my MQTT/NTP gateway being 0x01 throw that off? 

In the earlier timekeeping PR we said that if a peer was left as 0x00 (aka not defined) that we would not want to send the time to that address.  Hence I think we were assuming that 0x00 would never be assigned as an address to a legitimate gateway.  Therefore if the peer address is 0x00 I do not send the time to that address.

```
if((timeMaster.tmAddress != (ESPNOW1[4] << 8 | ESPNOW1[5])) && ESPNOW1[5] != 0x00) {
    DBG("Sending time to ESP-NOW Peer 1");
    result1 = sendESPNow(ESPNOW1, &sys_packet);
  }
```
```
if((timeMaster.tmAddress != (ESPNOW2[4] << 8 | ESPNOW2[5])) && ESPNOW2[5] != 0x00) {
    DBG("Sending time to ESP-NOW Peer 2");
    result2 = sendESPNow(ESPNOW2, &sys_packet);
  }

```

So if the ESP-NOW address of any of the peers ends in 0x00 we do not send the time to that peer.

timmbogner:
Ooooh, I see now. I was thinking there was an assumption that 0x00 was the MQTT gateway, sorry.  I don't have a better solution right now, except to send to the usually non-existent gateway.

This is only a faint idea, but what if the user could set `ESPNOW_NEIGHBOR_1` to "-1"? Then with a preprocessor condition we check if that value is negative, and we can set a flag in code or something so we know that gateway isn't real. That might be too obscure, but let me know what you think.

aviateur17:
I'll just remove the check for 0x00 and we can send to a non-existent peer. We can always change later if need be.

timmbogner:
> I'll just remove the check for 0x00 and we can send to a non-existent peer. We can always change later if need be.

Sorry, I just realized I worded that wrong. I should have said sending to non-existent peer was another option, not necessarily a better one. I don't think you need to change it as long as we are using 0x00 as a placeholder for a null (non-existent) gateway.

aviateur17:
> I added some debug messages and the transmissions began working. I believe I found the problem.
> 
> In `sendTimeLoRa()` you run `transmitLoRa()` twice in a row with no delay. If you are using `radio.startTransmit()`, you need to wait for the first operation to finish before doing it again. Probably just as easy to use `radio.transmit()`.
> 
> You can probably recreate the issue by disabling FDRS_DEBUG.

I tried radio.transmit last night and LoRa did not work.  I seem to recall from the switch over to RadioLib that my Hallard shields did not map one of the interrupts on like DIO2 to the chip.  I haven't looked into the details again recently and I don't know that's the issue but it could require more than one Interrupt.  I'll look in more detail later.

timmbogner:
I'll have to review the pins and interrupts and stuff when I'm back at my computer on Friday. To check into what is going on, I would insert `Serial.println("INT");` into the interrupt handler function, then look to see if and when the interrupt is being triggered. 
You should see it triggering after `startTransmit()` is finished, but if using `transmit()` you should see it triggered before the leaving the function. 

I'm sure we both know it's bad practice to do UART stuff in an ISR, but it works for this troubleshooting exercise. Probably don't try a DBG tho.

aviateur17:
Timm, thanks for the tip.  I will definitely try that.  Spent many hours last night trying to resolve a LoRa comm issue.  So frustrating in the number of times I've changed code and downloaded.  More troubleshooting to be done today.  To keep my sanity I may install the original code just to show myself it's not the hardware.

timmbogner:
I'm doing some research on my phone now. Yeah, I guess I forgot that the Hallard shield has NO interrupts mapped to GPIOs on the ESP. If that's the case then you might be just getting lucky in terms of timing thus far? If you are having sudden mysterious transmission/reception issues, also look at whether you have changed or removed DBGs. In the past i've found the timing held together only because of the delay caused by debugs. I'm not sure how it all can work for you without the interrupt, especially with my code that relies on it. I will review some more, but I was thinking that receiving was initiated by the interrupt being triggered.

Maybe try hardwiring an interrupt between the RFM95 and ESP to see if it helps?

aviateur17:
> I'm doing some research on my phone now. Yeah, I guess I forgot that the Hallard shield has NO interrupts mapped to GPIOs on the ESP. If that's the case then you might be just getting lucky in terms of timing thus far? If you are having sudden mysterious transmission/reception issues, also look at whether you have changed or removed DBGs. In the past i've found the timing held together only because of the delay caused by debugs. I'm not sure how it all can work for you without the interrupt, especially with my code that relies on it. I will review some more, but I was thinking that receiving was initiated by the interrupt being triggered.
> 
> Maybe try hardwiring an interrupt between the RFM95 and ESP to see if it helps?

DIO0, DIO1, and DIO2 on the LoRa Chip are connected to diodes and then to one GPIO PIn.  Depending upon how the chip is configured each DIO pin should generate an interrupt if something is happening.  Now how this works in the RadioLib software I'm not quite sure.  I'm not sure what the default configuration is for those DIO pins using the RadioLib software is.

timmbogner:
Oh that's right. Okay so maybe the solution for you is to remove the other diodes? In our usage on sx1278, Radiolib uses only the one interrupt (dio0). If we implement channel activity detection, that is where a second  interrupt comes into play (dio1). If your interrupt handler is being falsely triggered by dio1 instead of dio0, that could be an issue for you. Just a guess ...not sure if there are any signals happening on dio1 by default or not.

I don't have a clue what dio2 is doing, I'm not sure if it's even used.

timmbogner:
I'm somewhat keeping up with reading these, just haven't tested it in a bit.
08724ab is a PR of its own, but I see why you combined the files. Everything looks cool so far. I'm presently reading through your new DataReading release code. Could you transcribe how the txWindow stuff is working exactly? It doesn't look too complicated, I just want to have a good grasp on the concept.

timmbogner:
I have a small fix for the ping functionality in 'main' right now. Basically it prints the ping time before it records it, adding extra time to the returned time. This causes discrepancies in the displayed vs. returned time.

I know you've made a lot of ping changes here, is that even still an issue?

aviateur17:
> I'm somewhat keeping up with reading these, just haven't tested it in a bit. [08724ab](https://github.com/timmbogner/Farm-Data-Relay-System/commit/08724ab20deca060f36b50d4e3b6c8ce7288c420) is a PR of its own, but I see why you combined the files. 

Yeah, I was thinking the same thing.  It took quite a bit of time to do that.  I was having issues with LoRa transmissions and not receiving them on some gateways but was receiving them on others and after some banging my head and thinking I was going nuts I decided that I wanted to change to more of a Queue/Buffer instead of just transmitting right away.  I figured this might help with my issues - and I think it did.  If you would rather me do a separate PR on that I'd be glad to.  I'd probably start working on that before finishing this one though as, for me, it really seems to help my LoRa communication.

>Everything looks cool so far. I'm presently reading through your new DataReading release code. Could you transcribe how the txWindow stuff is working exactly? It doesn't look too complicated, I just want to have a good grasp on the concept.

The TxWindow stuff is really for retries or continuous transmissions so that there can be a little space in between transmissions for listening for responses.  With the buffer and the asynchronous nature, now there is no while loop to switch between transmit and receive and then wait awhile and back to transmit.   So this is a bit of a timer to say "wait for a little bit before trying to transmit again" otherwise for ACKs it would just transmit one after the other without taking time to listen for the ACK to come back.  The value could be adjusted to 0 to remove the functionality for testing.  I suppose the other option would be just specifically for ACKS to wait awhile for the re-transmissions but I also had an issue where if I perform a ping followed by a time request followed by some loads and sendFDRS then I wouldn't get a response from the ping because I'd be too busy transmitting all of that other stuff.  Now there some space to listen for a ping response before going on to the time request and then to the sendFDRS call.

aviateur17:
> I have a small fix for the ping functionality in 'main' right now. Basically it prints the ping time before it records it, adding extra time to the returned time. This causes discrepancies in the displayed vs. returned time.
> 
> I know you've made a lot of ping changes here, is that even still an issue?
> 
> EDIT: Looking over the code a little more, it looks like pingFDRS() doesn't return anything now. Is there still a way for the user to get/use the ping time?

Yeah, I implemented an asynchronous ping so it will just output the result to the screen.  What we could do, and Håkan's post gave me this idea, would be to implement a callback function when we get either a ping or a ping timeout and then we can send the ping time or lack of time to the callback function.  That way something can be done in response.  What do you think about that?

aviateur17:
> Is it possible to paste a timestamp into the serial console and set the time manually?

I've been thinking about this and it should be possible.  I'm just wondering how to prevent the input from being scrambled by all of the output that is happening.  Would be nice to pause the output for bit while waiting for input so the stuff on the screen doesn't get clobbered.  Any ideas on that?

aviateur17:
I think the functionality should be complete.  Let me know if you run into any issues.  Looks like there are merge conflicts so let me know if you want me to submit another PR or if you're going to manually merge the conflicts.  No rush on this as I know it's a lot of changes and everyone is busy.

aviateur17:
Thanks for catching that.  Here's what I'm thinking..

in fdrs_time.h line 130
```
bool validTime() {
  if(!VALID_TS(now) || timeSource.tmSource == TMS_NONE) {
    if(validTimeFlag) {
      DBG1("Time no longer reliable.");
      validTimeFlag = false;
    }
    return false;
.....
```
in the if() statement removing the check for " timeSource.tmSource == TMS_NONE"
What I was thinking about is that if the time hasn't been updated for a "long time" by some time source that the local time may not be valid or accurate.  My thought is to leave that check out and we can figure something out later if it becomes an issue.

timmbogner:
>I was having issues with LoRa transmissions and not receiving them on some gateways but was receiving them on others and after some banging my head and thinking I was going nuts I decided that I wanted to change to more of a Queue/Buffer instead of just transmitting right away. 

Circling back to this. I'm testing LoRa now, and I think the time is working. However there are a few DataReading features that got lost in this change (as far as I can tell). With the old functionality I was able to give the gateway more than a packet's worth (30-40+) of DRs and the asyncRelease would split them into several packets and send them to each destination sequentially. It seems your new functionality only supports a single LoRa packet to each recipient, which is not sufficient. 

Another behavior I noticed, but didn't look into closely yet was the way that DataReadings are sent via LoRa instantaneously upon arrival. The previous behavior was to save, combine and release them at a set interval. This was to limit radio chatter mostly, and to prepare to add channel activity detection (which would delay the release until the channel was clear). It also leaves it upon the user to set the delay in accordance with their local regulations regarding airtime and duty-cycle. I will have to look closer at your code. I guess I don't fully understand how your new buffering system works yet. Let me know if I'm misunderstanding things so far.
I've always considered that short SystemPackets, when sent individually, are an acceptable exception to that set transmission cycle. 
# 
`state_t` was in use by RadioLib so I changed it to `pingstate_t`.

timmbogner:
I guess this brings up a general question: Is it better to send a large amount of shorter packets, or a small amount of larger packets? Might be something to consult the Meshtastic community about. I'm definitely all ears if you have thoughts.

aviateur17:
> Circling back to this. I'm testing LoRa now, and I think the time is working. However there are a few DataReading features that got lost in this change (as far as I can tell). With the old functionality I was able to give the gateway more than a packet's worth (30-40+) of DRs and the asyncRelease would split them into several packets and send them to each destination sequentially. It seems your new functionality only supports a single LoRa packet to each recipient, which is not sufficient.

The code I've submitted does the same thing.  Basically it looks at all of the data in the queue and for all data that is going to the same destination address it sends that information together.

so in fdrs_lora.h line 830 we check if the buffer is not empty meaning there is data to be sent

line 837, if there is no existing data to be sent then we load new data to be sent
line 841, we load new data to be sent via malloc.  We know how many are to be sent via the previous call on line 839.  The transmitSameLoRa function looks at the queue and finds the number of DataReadings that are going to the same destination address and computs that number of DRs.  Now that we know the number we can allocate memory for that number of DataReadings on line 841.   Then, on line 843 to 845 in the for loop we populate the data to be sent via the for loop.  Take the data from the queue and mark it to be sent.  This can have up to len size DataReadings (more than 1 DataReading).  Then on line 846 we populate the address to be sent to.   Line 848 releases the data from the queue or buffer.   Now, depending upon whether we want an ACK or not, we send the data (len size DataReadings) via line 856 (no ACK) or line 866 (ACKs).

I've tested the functionality (via the LoRa Stress test sketch that is included) and I can see that it will send upwards of 10+ DataReadings at the same time.


> Another behavior I noticed, but didn't look into closely yet was the way that DataReadings are sent via LoRa instantaneously upon arrival. The previous behavior was to save, combine and release them at a set interval. This was to limit radio chatter mostly, and to prepare to add channel activity detection (which would delay the release until the channel was clear). It also leaves it upon the user to set the delay in accordance with their local regulations regarding airtime and duty-cycle. I will have to look closer at your code. I guess I don't fully understand how your new buffering system works yet. Let me know if I'm misunderstanding things so far. 

The TxDelay does this as well.  If there are more than 1 DataReadings that accrue in that delay time and they are all going to the same destination then the PR code will group them all together and attempt to send them at the same time.  

The delay could be increased, if need be, to accommodate regulatory requirements.  The longer the TxDelay the more time between transmissions.



>I've always considered that short SystemPackets, when sent individually, are an acceptable exception to that set transmission cycle.

I attempted this - excluding the SPs from the delay - and found that when I transmit a ping followed by a time request followed by DRs there is not enough receive time to reliably get the ping response.   Therefore, because of that testing, I decided that there should be some airtime between SP transmits as well.

> 
> `state_t` was in use by RadioLib so I changed it to `pingstate_t`.

Thanks!

aviateur17:
> I guess this brings up a general question: Is it better to send a large amount of shorter packets, or a small amount of larger packets? Might be something to consult the Meshtastic community about. I'm definitely all ears if you have thoughts.

In my mind this depends upon the regulatory requirements (laws in the area), how reliable longer transmissions are as compared to shorter transmissions, and what, if any, loss in control this has on the micro controller.  For the latter, the longer the transmits are the more execution needs to be done when the transmission completes.  The more features that are added to the project  the more critical it will be to "service" those features in a more timely and regular manner.   While I was doing that stress testing it seems like 10 DRs was taking about 200ms of transmission time.  It doesn't seem like that would be prohibitive at this point.

aviateur17:
> I think a point we're differing on is how many DRs is "a lot". As users scale up their systems, there could be a load of 30+ DataReadings arriving at a gateway at a time. Currently, I simulate this by inserting 30 JSON DataReadings into the console of 0x01, and send them via ESP-NOW to an 0x02 ESPNOW/LoRa gateway. 

Sounds good, so I think the first thing to do then is make the queue/buffer larger.  In fdrs_lora.h line 14 I make the buffer size 250/7 = 35.  We can definitely make that larger to something like 70 pretty easily by changing that line 14.   This will clear up the buffer full messages.

Then we have the issue of limiting the amount of LoRa data transmitted to the maximum data size of the LoRa packets.  I believe this is around 255 bytes or about 35 DataReadings, if I recall correctly.  I don't believe there is any limitation on the number of DRs to be sent (size is len) so we should probably add some code to check for the upper bound of len > 35 so that we only send 35 DataReadings at one time.

I guess this could be done in the transmitSameAddrLoRa() function (fdrs_lora.h line 380) at the return statement:
```
if(count > (250/sizeof(DataReading)) {
        return (250/sizeof(DataReading));
    }
    else {
        return count;
    }
```

timmbogner:
Okay I just realized ESP-NOW actually isn't splitting large packets correctly either. I'll need to look at that. 

BUT 😅 If on the main branch you go directly from serial to LoRa, and you paste in 36+ DataReadings, you will see it split the data into two packets and send them sequentially as follows. This behavior needs to be kept up. I'll look into ESP-NOW.



```
10:42:33.065 ->     Incoming Serial.
10:42:33.065 ->     Sending to LoRa broadcast buffer
10:42:37.241 ->     Transmitting LoRa message of size 251 bytes with CRC 0xea6e to LoRa MAC 0xffff
10:42:37.618 ->     Transmitting LoRa message of size 13 bytes with CRC 0xea8f to LoRa MAC 0xffff
10:42:37.658 ->     LoRa airtime: 440ms
```

aviateur17:
> Okay I just realized ESP-NOW actually isn't splitting large packets correctly either. I'll need to look at that.

I believe ESP-NOW does handle this.
fdrs_gateway_espnow.h line 408:
```
if(ln > espnow_size) {
```
This limits the transmission size to maximum ESP-NOW data transmission size.

> This behavior needs to be kept up.

Yep, for sure

timmbogner:
>I believe ESP-NOW does handle this.

Ah, I was still testing in the main branch, so I guess you fixed it. Thanks! I'm still digesting the changes to ESP-NOW, so bear with me. 

I'm going to commit the adjustments I have made so far, then continue testing the actual time portion.

timmbogner:
> Yeah, I implemented an asynchronous ping so it will just output the result to the screen. What we could do, and Håkan's post gave me this idea, would be to implement a callback function when we get either a ping or a ping timeout and then we can send the ping time or lack of time to the callback function. That way something can be done in response. What do you think about that?

Another circle-back:
Not sure if this is what you are describing... Would it be feasible to make a blocking ping function, just for the user to have access to, and it simply yields and runs the loop tasks until it gets a ping response or times out? Looking at the code, `recvPingEspNow(incMAC);` is run inside of the ESP-NOW callback. If I'm not mistaken, that means as long as you yield to the system, it will be executed as soon as a ping response comes in. Can we have it set a flag in `recvPingEspNow()` and `pingRequestLoRa()`, then wait for that flag at the end of the node's `pingFDRS()`? Also instead of clearing the `espNowPing` and `loraPing` structures, leave them intact and use the response time as the return or `pingFDRS()`.

Prototype:
``` cpp
int pingFDRS(uint32_t timeout)
{
  global_ping_flag = false;
#ifdef USE_ESPNOW
  pingFDRSEspNow(gatewayAddress, timeout);
#endif
#ifdef USE_LORA
  pingRequestLoRa(gtwyAddress, timeout);
#endif
while (!global_ping_flag){
  yield();
  handleLoRa();
}
 // do some stuff to get the response time
  return response_time;
}
```
There will need to be something add to handle timeouts, but I think this illustrates what I'm hoping to do.

aviateur17:
> > Yeah, I implemented an asynchronous ping so it will just output the result to the screen. What we could do, and Håkan's post gave me this idea, would be to implement a callback function when we get either a ping or a ping timeout and then we can send the ping time or lack of time to the callback function. That way something can be done in response. What do you think about that?
> 
> Another circle-back: Not sure if this is what you are describing... Would it be feasible to make a blocking ping function, just for the user to have access to, and it simply yields and runs the loop tasks until it gets a ping response or times out? Looking at the code, `recvPingEspNow(incMAC);` is run inside of the ESP-NOW callback. If I'm not mistaken, that means as long as you yield to the system, it will be executed as soon as a ping response comes in. Can we have it set a flag in `recvPingEspNow()` and `pingRequestLoRa()`, then wait for that flag at the end of the node's `pingFDRS()`? Also instead of clearing the `espNowPing` and `loraPing` structures, leave them intact and use the response time as the return of `pingFDRS()`.
> 
> Prototype:
> 
> ```c++
> int pingFDRS(uint32_t timeout)
> {
>   global_ping_flag = false;
> #ifdef USE_ESPNOW
>   pingFDRSEspNow(gatewayAddress, timeout);
> #endif
> #ifdef USE_LORA
>   pingRequestLoRa(gtwyAddress, timeout);
> #endif
> while (!global_ping_flag){
>   yield();
>   handleLoRa();
> }
>  // do some stuff to get the response time
>   return response_time;
> }
> ```
> 
> There will need to be something add to handle timeouts, but I think this illustrates what I'm hoping to do.

I wrote the following over several hours so it may be a little jumbled up...

I think the issue is not the ending part of the synchronous or blocking function it's the beginning part of it.  Once it's called we can execute it without issue it's just when calling it we need to make sure there is no async function already running.   That's why, in the LoRa ping code, I check to see if we are waiting for an ACK and if so, instead of running the ping right away, I queue up the ping for later execution.  If we didn't check for that we might miss the ACK and the ping response both.  If we ran the following code in a node..
```
pingFDRS();
reqTimeFDRS();
subscribeFDRS();
loadFDRS();
loadFDRS();
sendFDRS();
```
There are at least 4 LoRa transmits there.  When in that order there should be no issues running the synchronous code, but what if we ran it in this order.
```
loadFDRS();
loadFDRS();
sendFDRS();
pingFDRS();
subscribeFDRS();
reqTimeFDRS();
```
If everything was synchronous and we are using ACKS then we may miss the ACKS during the ping, subscribe, and time request.

The challenge when using both synchronous (blocking) and asynchronous functions is that when you call the synchronous function it may not be able to do what it needs to do right away due to an asynchronous function already in process.  So, for example, in LoRA, the transmissions of the  DataReadings  when using ACKS are asynchronous (if they were synchronous they may block for like 1 second or so).  If they are ongoing and then there is a call to a ping which is synchronous then we either cannot execute the ping or we have to wait and execute it later after the asynchronous task is completed.  With LoRa this is only complicated when ACKs are enabled as those require listening for responses and then potentially transmitting again.  So what I do in the ping and time request functions is to check to see if an Asynchronous task is running and if so I would queue up the ping for transmission later.  If you call this ping function that could be asynchronous there is no guarantee that it will return a value.   

What do you think about a callback function where when the ping completes, a function is called that can print to the output the ping response time and also hand that value to other code if you want to do something else with that value?   I see what information you desire but I'm not clear why it's needed as a response to pingFDRS and a callback function can't be used.  It might help if you could clarify what you're going to use the ping response time for.  I think pingFDRS returning a value makes everything clean and easy to understand.   Maybe what we could do is return a response value when the ping is performed synchronously and then just return 0 or some very high number when the ping is performed asynchronously.  

>Looking at the code, `recvPingEspNow(incMAC);` is run inside of the ESP-NOW callback. If I'm not mistaken, that means as long as you yield to the system, it will be executed as soon as a ping response comes in

I think that's the magic of using ESP-NOW and the second core. I think it generates an interrupt saying that there is a ESP-NOW response and then calls the callback function therefore that works when you're spinning in a while loop waiting for a response. 

>Can we have it set a flag in `recvPingEspNow()` and `pingRequestLoRa()`, then wait for that flag at the end of the node's `pingFDRS()`?

If you're spinning in a while loop in pingFDRS waiting for a response I don't think there's any other way to execute other code in general and wait for a response at the same time.   

I think a callback function would work but I'm open to doing what you think.

timmbogner:
I've realized that this ping stuff is a relatively minor feature, and the main reason I want to keep it is because it already exists in the old version. In reality, I must admit that being able to actually use the ping time is a fringe-feature. Few people will need it, although it is nice to have. I think the callback is just too much extra stuff for too little real benefit.
I also finally understand why the blocking/async thing is a problem. Thanks for explaining it so clearly. My method could work if there was a way to like, return -1 if there was error sending (it would fail if there was a transmission in progress). Even then though, I think it may be more trouble than it's worth.

So basically, I don't care too much about keeping the ability to use the ping time. Displaying is what most users will need, if they even use the ping functionality in the first place. You are welcome to implement your callback idea if you want, as long as it doesn't add much config overhead. Sorry for flip-flopping here!

aviateur17:
Sounds good, I'll fix up the ping stuff today (making them return int - both ESP-NOW and LoRa) and then do some more testing.

aviateur17:
Working on web page to set time, change debug level dynamically at runtime, and show debug output.

image

timmbogner:
I really like the webpage thing. Did you have to change DBG level from a macro to be configurable after compile?
I've discussed a SystemPacket command that would drop any gateway into WiFi mode for OTA programming. If we changed configurations to be settable after compile, we could invoke a captive portal to a page like yours with full configuration options. Configuration-in-the-field is an ages old request for FDRS.
Let's keep it simple for this PR, but I like the HTML backend concept a lot. 

I think I should have time for testing in the next week. After that I'll be on the road until roughly the end of the month.

aviateur17:
> I really like the webpage thing. Did you have to change DBG level from a macro to be configurable after compile? I've discussed a SystemPacket command that would drop any gateway into WiFi mode for OTA programming. 

I was thinking about that as well.  Would be a lot of work but this makes it easier. 

>If we changed configurations to be settable after compile (stored in flash/eeprom), we could invoke a captive portal to a page like yours with full configuration options. Configuration-in-the-field is an ages old request for FDRS. Let's keep it simple for this PR, but I like the HTML backend concept a lot.

I think I Just wanted to show you what I'm up to next.  I think I'd rather be done with this PR and keep this to the next PR.

> 
> I think I should have time for testing in the next week. After that I'll be on the road until roughly the end of the month.

Sounds good, no rush.  I don't think I'm planning on submitting any more to this PR unless I find bugs.

timmbogner:
I'm still here! When I left off a few weeks ago, the gateway functionality was tested with no issues. I will be testing nodes in the next week or so.

timmbogner:
@aviateur17 I finally solved the conflicts; maybe even correctly! It wasn't as painful as I'd anticipated. Please check over things to make sure I didn't unleash any unanticipated chaos 😅

The last thing I want to do is create a simple example of how to use the time data. The T-Watch sketch is a bit convoluted. I'm thinking just something that delivers each element of the time individually over the serial monitor. This can act as a framework that a user can adapt to a TFT/7-segment/OLED or whatever kind of clock display they desire to build.

If you have anything like that ready to go, or you'd like to contribute this, obviously please go for it! I plan to do it sometime, but I am not going to delay this PR any further either way. As long as this merge didn't break anything, I think the PR is ready to go!

aviateur17:
Timm, sounds good.  It may take a couple of weeks as I don't have a lot of time right now but I'll check it out and make comments to keep up to date.

aviateur17:
Output from Controller Time example to be submitted shortly:

The timestamps in front are my personal added code so won't be shown.

```
20422 |     FDRS ControllerTime example.
[ 30464][W][Wire.cpp:301] begin(): Bus already started in Master Mode.
[ 30464][E][esp32-hal-gpio.c:102] __pinMode(): Invalid pin selected
E (30462) gpio: gpio_set_level(227): GPIO output gpio_num error
E (30497) gpio: gpio_set_level(227): GPIO output gpio_num error
[ 30500][W][Wire.cpp:301] begin(): Bus already started in Master Mode.
30530 |     Display initialized!
30554 |     Hello, World!
30595 |     FDRS User Node initializing...
30636 |      Reading ID 85
30676 |      Gateway: 21
30717 |     Initializing ESP-NOW!
30757 |     Failed to add peer bcast
30798 |     Requesting time from gateway
30838 | [1] Requesting time from gateway 0x21
30845 | [2] Incoming ESP-NOW System Packet from 0x21
30845 | [1] Received time via ESP-NOW from 0x21
30846 | [1] ESP-NOW time source is now 0x21
30856 | [1] Time adjust 1714575195 secs
30856 | [1] Time change from STD -> DST
1714575195 |     ---- strftime() function output ----
1714575195 |     Local date/time is: Wed May  1 09:53:15 2024 DST
1714575195 |     Year: 2024
1714575195 |     Month: 05
1714575195 |     Month Name: May
1714575195 |     Day of Month: 01
1714575195 |     Day of Week: 3
1714575195 |     Day Name: Wednesday
1714575195 |     Day of Year: 122
1714575195 |     Hour (24 hour format): 09
1714575195 |     Hour (12 hour format): 09
1714575195 |     Minute: 53
1714575195 |     Second: 15
1714575195 |     AM/PM: AM
1714575195 |     US format: Wednesday, 05/01/2024 09:53:15 AM
1714575195 |     European format: Wednesday, 01/05/2024 09:53:15
1714575195 |     ---- struct tm output ----
1714575195 |     Year: 2024
1714575195 |     Month: 5
1714575195 |     Month Name: May
1714575195 |     Day of Month: 1
1714575195 |     Day of Week: 3
1714575195 |     Day Name: Wednesday
1714575195 |     Day of Year: 122
1714575195 |     Hour (24 hour format):9
1714575195 |     Hour (12 hour format):9 AM
1714575195 |     Minute: 53
1714575195 |     Second: 16
1714575195 |     Local Time: Wednesday, 5/1/2024 9:53:16 AM
1714575223 |     Local date/time is: Wed May  1 09:53:44 2024 DST
```

aviateur17:
Timm, the merge seems to be good based on my review of your commit.  I think there may be an issue from my doing in the node_espnow code in the pingEspNow function.  Let me review that and make sure everything compiles properly.

aviateur17:
Timm, I just made the changes mentioned earlier. I think we're good to go.

timmbogner:
I'm going to merge this (sorry it's taken so long), but I still need to add to the documentation. I'll get to that once my springtime work dies down.

I noticed that LilyGo has a new T-Watch with an integrated LoRa radio. I'm excited to order one, and I'd love to send you one too as a token of my thanks for everything you do for FDRS. Email me if you're interested!

PR #214, opened 2024-06-11
Optionally Synchronous sendFDRS()

timmbogner:
With this change, the system sends all buffered LoRa transmissions before leaving sendFDRS(). This emulates a synchronous send routine, despite our background operations now occuring asynchronously. If a user is sending and receiving data at the same time, they can use the asynchronous sendFDRS(true), which only sends the data to the buffer.

This still needs:

- the examples currently using sendFDRS() asyncronously to be changed,
- documentation adjusted
- testing that ESP-NOW isn't affected
- optimize deep-sleep. It needs to clear out or "flush" the send buffer before sleeping. (Thanks for reminding me of that word @fairoldi :D)

### DIFF
--- a/src/fdrs_node.h
+++ b/src/fdrs_node.h
@@ -220,6 +220,7 @@ bool sendFDRSinternal()
 bool sendFDRS(bool is_async)
 {
   bool ret = sendFDRSinternal();
+#ifdef USE_LORA
   if (!is_async)
   {
     unsigned long timeout = millis() + 1000;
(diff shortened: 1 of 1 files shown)


aviateur17:
I don't think this gonna quite work.    This makes sure the Async Queue is empty but doesn't make sure that the transmit is complete.  The Async queue (circular buffer) transfers into a temporary buffer (linear buffer) and then sends out.   If ACKs are enabled then we need to wait for them to come back from the gateway and may potentially need to wait for a couple of them.  So in an ideal situation this may work but I'd like to take another crack at it.  I hope that's okay.    I'm thinking about bringing back old transmitLoRa function code that was synchronous and then we can default sendFDRS to use that synchronous transmitLoRa function.

PR #216, opened 2024-06-13
Mqtt topics

aviateur17:
Created Global Defines for MQTT Topics:

// MQTT Topics
#define GLOBAL_TOPIC_DATA    "fdrs/data"
#define GLOBAL_TOPIC_STATUS  "fdrs/status"
#define GLOBAL_TOPIC_COMMAND "fdrs/command"
#define GLOBAL_TOPIC_DATA_BACKLOG "fdrs/databacklog"   // Used in filesystem module


timmbogner:
This looks great! I'm not sure if we're going to need to add all the new options to the example sketch configs. My idea: I think we should start a list of configs that the user can define. Then we can add a quick bit of documentation to the config file about how local vs. global configs work, as well as the list of configs that can be defined locally. Then I'd like to condense the configs a bit and  remove seldom-used options like the MQTT topics and static IP configs, plus move the WiFi credentials and MQTT info closer to the top. I'll think about the blurb and try to get those changes submitted in the next couple days.

This also has the irrigation example changes in it, which I had one last note on: You have the `registerWithGateway()` function that's essentially doing what already should be done in the background by the ESP-NOW controller's `refresh_registration()` function. To my knowledge the only device that should need to register are ESP-NOW controllers, since LoRa controllers just process all incoming broadcast packets from their gateway. If this is just an oversight, no problem. If there's something new with LoRa timekeeping requiring more connection maintenance, let me know.

The other stuff sounds good too, but I haven't had a chance to dive in. I kinda only have time for this one response this morning, but I'll make the rounds later on.

aviateur17:
> This looks great! I'm not sure if we're going to need to add all the new options to the example sketch configs. My idea: I think we should start a list of configs that the user can define. Then we can add a quick bit of documentation to the config file about how local vs. global configs work, as well as the list of configs that can be defined locally. Then I'd like to condense the configs a bit and remove seldom-used options like the MQTT topics and static IP configs (they'll be moved to the list), plus move the WiFi credentials and MQTT info closer to the top. I'll think about the blurb and try to get those changes submitted in the next couple days.

Maybe we can split that out into a discussion topic - User definable configuration options

> 
> This also has the irrigation example changes in it, which I had one last note on: You have the `registerWithGateway()` function that's essentially doing what already should be done in the background by the ESP-NOW controller's `refresh_registration()` function. To my knowledge the only device that should need to register are ESP-NOW controllers, since LoRa controllers just process all incoming broadcast packets from their gateway. If this is just an oversight, no problem. If there's something new with LoRa timekeeping requiring more connection maintenance, let me know.

I did not intend to include that example in there.  Not sure if I can go back and change it but I will try to back out that change later on.   I'll copy this comment into that specifc PR and we can address that in the discussion related to that irrigation sketch PR.

timmbogner:
Two things presently:
- The scheduler changes are still in this, but that's maybe okay. I can pull this with the scheduler and and then you can resubmit #205 with only the sketch changed. Is that cool?
- I don't think the backlog topic is used anymore and can be removed. With no objections, I'll do that tonight-ish in this PR, then pull it.

aviateur17:
> Two things presently:
> 
> * The scheduler changes are still in this, but that's maybe okay. I can pull this with the scheduler and and then you can resubmit [Update irrigation example to support time limited activation, small c… #205](https://github.com/timmbogner/Farm-Data-Relay-System/pull/205) with only the sketch changed. Is that cool?
> * I don't think the backlog topic is used anymore and can be removed. With no objections, I'll do that tonight-ish in this PR, then pull it.

Done

timmbogner:
Thanks!

PR #217, opened 2024-06-13
revert sendFDRS to Synchronous operations, add sendFDRSAsync

aviateur17:
Initial coding for Synchronous sendFDRS()

Need to test compilation and functionality before this PR is complete.

### DIFF
--- a/src/fdrs_lora.h
+++ b/src/fdrs_lora.h
@@ -140,7 +140,9 @@ volatile bool operationDone = false;  // flag to indicate that a packet was sent
 unsigned long loraAckTimeout = 0;
 unsigned long rxCountDR = 0;              // Number of total LoRa DR packets destined for us and of valid size
 unsigned long rxCountSP = 0;               // Number of total LoRa SP packets destined for us and of valid size
+unsigned long rxCountUnk = 0;               // Number of total LoRa packets received but not known (invalid size or format)
 unsigned long rxCountCrcOk = 0;             // Number of total Lora packets with valid CRC
+unsigned long rxCountCrcBad = 0;             // Number of total Lora packets with valid CRC
 unsigned long txCountDR = 0;                // Number of total LoRa DR packets transmitted
 unsigned long txCountSP = 0;                // Number of total LoRa SP packets transmitted
 extern time_t now;
@@ -195,7 +197,7 @@ void printLoraPacket(uint8_t *p, int size)
   printf("Printing packet of size %d.", size);
   for (int i = 0; i < size; i++)
   {
-    if (i % 2 == 0)
+    if (i % 16 == 0)
       printf("\n%02d: ", i);
     printf("%02X ", p[i]);
   }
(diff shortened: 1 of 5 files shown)


aviateur17:
Calls to sendFDRS from nodes will be synchronous.  Added sendFDRSAsync() to use Asynchronous calls.   Added code in sleepFDRS to check for Async functionality complete before going to sleep.

src/fdrs_lora.h
```diff
@@ -286,6 +289,52 @@ void transmitLoRa(uint16_t *destMac, SystemPacket *packet, uint8_t len)
     return;
 }
 
+// Synchronously Transmits LoRa data and, if configured, listens for ACKs
+// Return type is crcResult struct - CRC_OK, CRC_BAD, CRC_NULL.  CRC_NULL used for non-ack data
+crcResult transmitLoRaSync(uint16_t *destMac, DataReading *packet, uint8_t len)
+{
+    crcResult crcReturned = CRC_NULL;
+    // Do we check to see if another Async operation is in progress first???
+    // If Yes, what do we do if another operation is in progress?  Return and rely on retransmit retry?
```
timmbogner, line 298: I think that if an operation is in progress then we should either:
- Wait until it finishes and queue it after. I think this is probably messy and non-ideal.
- Fail. It would be nice to return -1 for 'busy', then maybe the user could do `while(sendFDRS() = -1){delay(100);}` to retry. That's a bit extra though. In the end, the user should probably just be careful using both types in the same sketch. I guess there's also the possibility of a receive op happening.

I would say I'm fine with just returning false (from `sendFDRS()`) if there's an op in progress.

src/fdrs_node.h
```diff
@@ -194,27 +196,49 @@ bool sendFDRS()
...
+
+// Sends FDRS data asynchronously - no spinning loop, more processor friendly, but does not return as much information
+// Returns true if data is queued up, false if there is no data to be sent
+bool sendFDRSAsync()
+{
+  if(data_count == 0) {
+    return false;
+  }
+
+  DBG("Queueing FDRS Packet!");
```
timmbogner, line 231: Great idea on the terminology change

aviateur17:
Working through issue where large amount of DR Data (~40 DRs) sent via LoRa the receiving gateway is getting garbage data and CRC Mismatch.  Up to 32 DRs has no issues. 

Also, when splitting large amount of DR data (~40 DRs) into multiple transmits (packet size exceeds 250 bytes), receiver is not receiving 2nd set of data.

timmbogner:
Are these issues affecting all operations or just synchronous ones?

We've run into issues in LoRa stuff that required arbitrary delays before, but 500ms is a lot. I'd like to study up on where/why the timing is getting thrown off before we do an algorithm to make up for it. It's worth noting that I've needed to add delays for compatibility with SX1262 in the past. 

Just saw edit 3:
That seems like the best fix for now, but there has to be a better way. I might try a deep-dive into the LoRa timing stuff soon, as I'll hopefully have some free time on an upcoming vacation. I think we can add the synchronous LoRa stuff sooner, and then fix the large transmission issue.

aviateur17:
> Just saw edit 3: That seems like the best fix for now, but there has to be a better way. I might try a deep-dive into the LoRa timing stuff soon, as I'll hopefully have some free time on an upcoming vacation. I think we can add the synchronous LoRa stuff sooner, and then fix the large transmission issue.

Agreed.  We can handle the larger data operations in another PR.  Hopefully can submit this soon.

aviateur17:
Got the inter-message delay down to 100ms and large number of DRs seem to work properly with and without ACKs.  Creating commits soon.

aviateur17:
I believe this should be complete.  Let me know if you run into or see any issues.

PR #218, opened 2024-06-13
Timekeeping defines

aviateur17:
Add defines in fdrs_global.h for USDST and EUDST and related timekeeping defines. Updated config files for those new defines.


timmbogner:
I read that the EU calls it "summer time". I clarified just in case "DST" isn't as common elsewhere.

PR #220, opened 2024-06-20
mass data type added

Stips5:


Stips5:
added data types for mass measurements

timmbogner:
Awesome, thanks for cleaning it up a bit too!

PR #227, opened 2024-06-29
Add leading hex zero to addresses

timmbogner:
@aviateur17 Hey so this is something that has stumped us both before, but your irrigation code gave me an idea. Here's the proof of concept I wrote:
``` cpp
  Serial.begin(115200);
  uint8_t x = 2;
  Serial.println("It's 0x" + String((x < 16) ? "0" : "") + String(x, HEX));
```

I applied it to the gateway's ESP-NOW functions and it seems to work. However, when I started expanding it further I realized there are a lot of spots where we'd use this, especially in LoRa. I think we can probably transform it into a reusable function to keep things neat. What do you think?

### DIFF
--- a/src/fdrs_gateway_espnow.h
+++ b/src/fdrs_gateway_espnow.h
@@ -48,14 +48,14 @@ void OnDataRecv(const esp_now_recv_info *pkt_info, const uint8_t *incomingData, 
   #endif
   if (len < sizeof(DataReading))
   {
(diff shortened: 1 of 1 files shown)


aviateur17:
That looks excellent.  We could use a macro like the ones at the top of fdrs_lora.h
```
#define LZ(a) (" 0x" + String(a < 16) ? "0" : "") + String(a, HEX))
```

Something like that and then just insert LZ(5) or whatever in the code
Great idea!

Issue #233, opened 2024-08-28
unable to compile ...


urs8000:
in case of missing <sys/time.h>   board selected: Feather M0 as a *sensor* which does not use time
the above mentioned library is only valid for ESP32
why don't you use an universal time library for that functions?
e.g. Time with TimeLib.h ???

timmbogner:
Hi @urs8000,
There are issues with AVR boards not having the [sys/time.h](https://pubs.opengroup.org/onlinepubs/007908775/xsh/systime.h.html) file, but (to my knowledge) it should be present for SAMD and other boards/chips with time functionality. I am testing the 'LoRa_Sensor' sketch under the _Adafruit Feather M0 (SAMD21)_ definition and it compiles successfully. Let me know if your setup is different.

urs8000:
hello @timmbogner 
thank for your answer
I installed FDRS from scratch and could compile your unmodified example LoRa_Sensor without error (only some notes & warnings)
as soon as I activate DGB_Level 1 or 2 I get:
In file included from e:\_development\Arduino\libraries\Farm-Data-Relay-System\src/fdrs_node.h:40,
                 from E:\_development\Arduino\libraries\Farm-Data-Relay-System\examples\LoRa_Sensor\LoRa_Sensor.ino:10:
e:\_development\Arduino\libraries\Farm-Data-Relay-System\src/fdrs_time.h: In function 'void adjTimeforNetDelay(time_t)':
e:\_development\Arduino\libraries\Farm-Data-Relay-System\src/fdrs_time.h:391:45: error: call of overloaded 'String(time_t&)' is ambiguous
  391 |       DBG1("Time adj by " + String(newOffset) + " secs");

at the end I could compile HELTEC_LoRa_32(V2), Feather_M0(SAMD21)_LoRa and other ESP based controllers
but NOT ANY ARM based e.g. MoteinoMEGA(-USB)

to overcome this problems I've used a save library bundle dated of 26/02/2024 to compile onla the MoteinoMEGA-sensor

it would be great if you could EXCLUDE timeinformation from any sensor. that's a function one (best MQTT) of the gateways should fullfill.

urs8000:
hello @timmbogner 
thank for your answer
I installed FDRS from scratch and could compile your unmodified example LoRa_Sensor without error (only some notes & warnings)
as soon as I activate DGB_Level 1 or 2 I get:
[...]

PR #238, opened 2024-12-09
AVR compatibility with LoRa Node and time functions

aviateur17:
Allow use of AVR boards as LoRa nodes per issue #233.

### DIFF
--- a/src/fdrs_datatypes.h
+++ b/src/fdrs_datatypes.h
@@ -1,7 +1,11 @@
 // A list of all datatypes you can use within FDRS.
 // If you are missing any data type, please open an issue at:
 // https://github.com/timmbogner/Farm-Data-Relay-System/issues
- 
+
+#ifdef AVR
+typedef unsigned int uint;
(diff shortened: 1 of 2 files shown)


timmbogner:
Kewl!
Interestingly, while the sensor example is slightly too big to compile under 328P or 32U4, it is almost small enough when it is stripped down to only the library. For example the following sketch (with standard LoRa_Sensor config file) actually does compile for Uno:
``` cpp 
#include "fdrs_node_config.h"
#include <fdrs_node.h>

void setup() {
  beginFDRS();
}
void loop() {
}
```
```
Sketch uses 26402 bytes (81%) of program storage space. Maximum is 32256 bytes.
Global variables use 980 bytes (47%) of dynamic memory, leaving 1068 bytes for local variables. Maximum is 2048 bytes.
```

As soon as you add `sendFDRS()` it goes over the limit though, so probably not really worth fighting for. The bigger AVR chips now compile with room to spare!

PR #240, opened 2025-02-09
Add NOISE as a data type

marius-cornescu:
Adding noise as a data type


marius-cornescu:
> For reference (and my curiosity), is this to measure noise in a signal or audible noise? What's your use-case?

It is for audible noise / dB(A)

PR #244, opened 2025-03-02
fix: type conversion and remove deprecated ArduinoJson 7.2 function

sivar2311:
fix: deprecated ArduinoJson 7.2 function `containsKey()` -> `is<dataType>()`
fix: type conversion `time_t` -> `uint32_t`


timmbogner:
Thanks!

